PR #1, opened 2018-05-07
Initial implementation

aknysh:
## what
* Terraform generic modules to provision common AWS resources
* A collection of reusable root modules for CloudPosse AWS accounts

## why
* To be reused across different CloudPosse AWS accounts (`prod`, `staging`, `dev`, `audit`)

### DIFF
--- a/aws/backing-services/aurora-postgres.tf
+++ b/aws/backing-services/aurora-postgres.tf
@@ -1,0 +1,77 @@
+# Don't use `admin`
+# ("MasterUsername admin cannot be used as it is a reserved word used by the engine")
+variable "POSTGRES_ADMIN_NAME" {
+  type        = "string"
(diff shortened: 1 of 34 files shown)


aws/backing-services/aurora-postgres.tf
```diff
@@ -0,0 +1,77 @@
...
+}
+
+variable "POSTGRES_DB_NAME" {
+  type        = "string"
+  description = "Postgres database name"
+}
+
+variable "POSTGRES_INSTANCE_TYPE" {
+  type        = "string"
+  default     = "db.r4.large"
```
sarkis: Should we default to something smaller here for the example? I worry people use this later and incur charges they didn't expect :) - if they forget to clean up. Maybe: db.t2.small or db.t2.medium?

PR #8, opened 2018-06-04
Remove dependencies on `terraform-aws-account-metadata`. Use Terraform vars.

aknysh:
## what
* Remove dependencies on `terraform-aws-account-metadata`
* Use Terraform vars instead (which could be be provided from `Dockerfile` or `chamber`)

## why
* Make`terraform-root-modules` a generic Terraform module that could be used for different projects (including third-party) and for Cloud Posse documentation
* `terraform-aws-account-metadata` is a private repo and contains secrets
* Eliminate cold-start cycles
* Fewer repos to maintain


osterman:
I think this is a good change.

PR #9, opened 2018-06-05
Add `organization` project

aknysh:
## what
* Add `organization` project

## why
* To easily create an AWS Organization from the master account

### DIFF
--- a/aws/organization/main.tf
+++ b/aws/organization/main.tf
@@ -1,0 +1,48 @@
+# NOTE: Organization can only be created from the master account
+# https://www.terraform.io/docs/providers/aws/r/organizations_organization.html
+
+terraform {
+  required_version = ">= 0.11.2"
+
+  backend "s3" {}
+}
+
+variable "aws_assume_role_arn" {
+  type = "string"
+}
+
+variable "organization_feature_set" {
+  type        = "string"
+  default     = "ALL"
+  description = "Specify `ALL` (default) or `CONSOLIDATED_BILLING`"
+}
+
+provider "aws" {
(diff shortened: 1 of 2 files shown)


osterman:
Help me understand how you will use this =)

aknysh:
* we want everything as code
* when we create a parent account in AWS, it does not have an organization - we need to create it
* we provision the org and all member accounts (`prod`, `staging`, `dev`) from TF instead of manually creating them from the console

So in the `root` `Dockerfile` we do:

```dockerfile
COPY --from=terraform-root-modules /aws/organization/ /conf/organization/
```

Although maybe we can move it to `root.xxxxx.xxx` because it will be used only in the `root`

osterman:
I'd like to move everything in `root.cloudposse.co` to this repo as well. I think everything should get copied from here.

osterman:
That said, how do you plan to use this for provisioning prod, dev, staging, etc. orgs? One project per org?

aknysh:
but the org is always only one.
it's metadata on top of the master account (root in our case).
`prod`, `staging`, `dev` are not orgs, they are member accounts.

1. We manually create the `parent/root` account from AWS console (no way to automate it)
2. From `root.xxxxx.xxx` `geodesic` shell we provision the org
3. From `root.xxxxx.xxx` `geodesic` shell we provision all member accounts (`prod`, `staging`, `dev`)

Member accounts can be organized into Organizational Units (OUs).
But Organization is always only one per parent (billing) account.

osterman:
Thanks! I was confused with `organization` vs `account`.

PR #10, opened 2018-06-08
Add projects from `root` stage. Add `account` project

aknysh:
## what
* Add projects (`dns` and `iam`) from `root` stage (and  eliminate dependencies of those projects on `terraform-account-metadata`)
* Add `accounts` project

## why
* Keep all reference architecture modules in one repo
* Provision member accounts (`prod`, `staging`, `dev`, `audit`) in an organization


aws/account/terraform.tfvars.example
```diff
@@ -0,0 +1,11 @@
+prod_account_name="prod"
+prod_account_email="info+prod@cloudposse.co"
+
+staging_account_name="staging"
```
osterman: I think we should maybe separate them into N files.

image

aws/dns/stages.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_route53_zone" "dns_zone" {
```
osterman: Why is this file called `stages.tf`?

aws/iam/audit.tf
```diff
@@ -0,0 +1,14 @@
...
+  description = "Audit account ID"
+}
+
+# Provision group access to audit account. Careful! Very few people, if any should have access to this account.
+module "organization_access_group_audit" {
+  source            = "git::https://github.com/cloudposse/terraform-aws-organization-access-group.git?ref=tags/0.1.2"
+  namespace         = "${var.namespace}"
+  stage             = "audit"
+  name              = "admin"
+  user_names        = ["erik", "andriy", "igor", "sarkis"]
```
osterman: Let's switch to email addresses. Move to `terraform.tfvars`

aws/iam/prod.tf
```diff
@@ -0,0 +1,14 @@
...
+  description = "Prod account ID"
+}
+
+# Provision group access to production account
+module "organization_access_group_prod" {
+  source            = "git::https://github.com/cloudposse/terraform-aws-organization-access-group.git?ref=tags/0.1.2"
+  namespace         = "${var.namespace}"
+  stage             = "prod"
+  name              = "admin"
+  user_names        = ["erik", "andriy", "igor", "sarkis"]
```
osterman: move to var file

aws/iam/root.tf
```diff
@@ -0,0 +1,9 @@
+# Provision group access to root account with MFA
+module "organization_access_group_root" {
+  source           = "git::https://github.com/cloudposse/terraform-aws-iam-assumed-roles.git?ref=tags/0.2.0"
+  namespace        = "${var.namespace}"
+  stage            = "root"
+  admin_name       = "admin"
+  readonly_name    = "readonly"
+  admin_user_names = ["erik", "andriy", "igor", "sarkis"]
```
osterman: move to var file

aws/dns/stages.tf
aknysh: the file will be copied into the stages (prod, staging, dev) in `Dockerfiles`.
the file is re-used.
not sure what to call it better.

aknysh: maybe just `dns`?

aws/account/audit.tf
```diff
@@ -0,0 +1,21 @@
+variable "audit_account_name" {
+  type        = "string"
+  description = "Audit account name"
+  default     = "audit"
+}
+
```
osterman: Thinking `account/` should be `accounts/`

aws/dns/stages.tf
osterman: Ok, I think perhaps we are overloading `dns/`. 

Thinking we should have `root-dns` and maybe `account-dns`

aws/account/main.tf
```diff
@@ -0,0 +1,27 @@
...
+
+variable "account_role_name" {
+  type        = "string"
+  description = "IAM role that Organizations automatically preconfigures in the new member account"
+  default     = "OrganizationAccountAccessRole"
+}
+
+variable "account_iam_user_access_to_billing" {
+  type        = "string"
+  description = "If set to ALLOW, the new account enables IAM users to access account billing information if they have the required permissions. If set to DENY, then only the root user of the new account can access account billing information"
```
osterman: Use markdown code? 
```
If set to `ALLOW`
```

aws/dns/stages.tf
osterman: (open to alternative names too... but i think we should create more folders rather than less)

osterman: A given folder should never serve 2 purposes.

PR #15, opened 2018-06-11
Added install scripts

goruha:
## What
* Added shell script to install `tf-state` module
* Added shell script to install `root-iam` module

## Why
* To simplify installation of `tf-state`
* To simplify installation of `root-iam`


aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+sed -i "s/backend/#backend/" main.tf
```
aknysh: this will change `backend` in two other places:
https://github.com/cloudposse/terraform-root-modules/blob/master/aws/tfstate-backend/main.tf#L54
https://github.com/cloudposse/terraform-root-modules/blob/master/aws/tfstate-backend/main.tf#L55

I did this before:
```
sed -i 's/backend          "s3"             {}/#backend          "s3"             {}/' main.tf
sed -i 's/#backend          "s3"             {}/backend          "s3"             {}/' main.tf```
```

but it's too fragile and could be broken if the file changes
We need to think of a better way.

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,23 @@
...
+export TF_BUCKET_REGION=${TF_VAR_region}
+
+sed -i "s/#backend/backend/" main.tf
+
+s3 fstab "${TF_BUCKET}" '/' '/secrets/tf'
+
+echo "yes" | init-terraform
+
+
+echo "Add to the Geodesic Module Dockerfile following"
```
aknysh: Add to the Geodesic Module Dockerfile the following ENV vars:

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ backend / #backend /" main.tf
+sed -i "s/ role_arn / #role_arn /" main.tf
```
osterman: This can't be generally true. This is maybe true for the "root" org, but certainly not on subaccounts.

osterman: roles are how we determine the account in which to provision the resources, so if this is not set, terraform won't use the appropriate account.

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+
+init-terraform
+terraform plan
+
+export TF_BUCKET=$(echo "yes" | terraform apply | grep -o -e "tfstate_backend_s3_bucket_id\s=\s.*" | cut -d ' ' -f 3)
```
osterman: Clever, but use `terraform output` instead. 

```
terraform output -json | jq -r .tfstate_backend_s3_bucket_id.value
```

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+sed -i "s/ #backend / backend /" main.tf
+
+echo "yes" | init-terraform
+
+sed -i "s/ #role_arn / role_arn /" main.tf
+
+echo "Add to the Geodesic Module Dockerfile following"
+echo "#----------------------------------------------"
+echo "ENV TF_BUCKET=\"${TF_BUCKET}\""
+echo "ENV TF_BUCKET_REGION=\"${TF_BUCKET_REGION}\""
```
osterman: Output lock table too:
```
terraform output -json | jq -r .tfstate_backend_dynamodb_table_id.value
```

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+export TF_BUCKET_REGION=${TF_VAR_region}
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ #backend / backend /" main.tf
+
+echo "yes" | init-terraform
+
+sed -i "s/ #role_arn / role_arn /" main.tf
+
+echo "Add to the Geodesic Module Dockerfile following"
```
osterman: Add the following to the Geodesic Module's Dockerfile:

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ backend / #backend /" main.tf
```
osterman: Let's be more precise. 
```
sed -Ei 's/^(\s+backend\s+)/#\1/' main.tf
```

aws/tfstate-backend/install.sh
osterman: I think this will need to be an arg. e.g. `--disable-role-arn`

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+
+init-terraform
+terraform plan
+
+export TF_BUCKET=$(echo "yes" | terraform apply | grep -o -e "tfstate_backend_s3_bucket_id\s=\s.*" | cut -d ' ' -f 3)
+export TF_BUCKET_REGION=${TF_VAR_region}
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ #backend / backend /" main.tf
```
osterman: Use regex style from above.

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "yes" | init-terraform
+
+sed -i "s/ #role_arn / role_arn /" main.tf
```
osterman: Use regex style from above

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Add to the Geodesic Module Dockerfile following"
+echo "#----------------------------------------------"
+echo "ENV TF_BUCKET=\"${TF_BUCKET}\""
+echo "ENV TF_BUCKET_REGION=\"${TF_BUCKET_REGION}\""
+echo "#----------------------------------------------"
+echo "And rebuild the module"
```
osterman: Then rebuild the geodesic module.

aws/tfstate-backend/install.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ backend / #backend /" main.tf
+sed -i "s/ role_arn / #role_arn /" main.tf
+
+
+init-terraform
+terraform plan
+
```
osterman: Run `terraform apply` here.

aws/tfstate-backend/install.sh
aknysh: yes, we disable the role only in the `root` account and only once at cold start when we provision `iam` and we don't have any roles yet

aws/accounts/install.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -i "s/ role_arn / #role_arn /" main.tf
```
osterman: See previous PR comments. This was not updated.

aws/cloudtrail/install.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
osterman: I don't want this script duplicated into each project. If we need that, then move it to `geodesic`

aws/accounts/install.sh
```diff
@@ -0,0 +1,12 @@
...
+
+init-terraform
+terraform plan
```
tamsky: suggest perhaps:
```
init-terraform
terraform plan -input=false -out=install-plan &&
  terraform apply -input=false install-plan &&
  mv install-plan install-plan.applied
```

aws/tfstate-backend/install.sh
tamsky: So the way I typically do something like this, is:
have two different `make` targets with
two different "invisible-to-terraform™" versions of the resource definitions:
- `provider-with-assume-role.tf.in`
- `provider-bootstrap.tf.in`

and the `make` targets look like:
```
bootstrap:
    rm -f provider.tf
    ln -s provider-bootstrap.tf.in provider.tf

default:
   rm -f provider.tf
   ln -s provider-with-assume-role.tf.in provider.tf
```

and the install script then looks closer to:
```
init-terraform
make bootstrap
terraform plan -input=false -out=install-plan &&
  terraform apply -input=false install-plan &&
  mv -v install-plan install-plan.applied
```

osterman:
Fix PR name

PR #20, opened 2018-06-28
Added ecr

goruha:
## What
* ECR for kops

## Why
* Deploy Docker images to K8S from ECR

### DIFF
--- a/aws/ecr/ecr-repo-app.tf
+++ b/aws/ecr/ecr-repo-app.tf
@@ -1,12 +1,12 @@
 module "kops_ecr_app" {
-  source    = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.0"
-  namespace = "${var.namespace}"
-  stage     = "${var.stage}"
-  name      = "langing"
+  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.0"
+  namespace    = "${var.namespace}"
+  stage        = "${var.stage}"
+  name         = "${element(repositories_names,0)}"
   cluster_name = "${var.region}.${var.zone_name}"
 
   users = [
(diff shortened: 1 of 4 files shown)


aws/ecr/ecr-repo-app.tf
```diff
@@ -0,0 +1,40 @@
+module "kops_ecr_app" {
+  source    = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.0"
+  namespace = "${var.namespace}"
+  stage     = "${var.stage}"
+  name      = "langing"
```
osterman: Typo

osterman: make it a variable. Add `terraform.tfvars.example` with defaults.

aws/ecr/ecr-repo-app.tf
```diff
@@ -0,0 +1,40 @@
+module "kops_ecr_app" {
+  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.0"
+  namespace    = "${var.namespace}"
+  stage        = "${var.stage}"
+  name         = "${element(repositories_names,0)}"
```
osterman: This is an invalid variable :P

osterman: Also, I don't understand why it's valid that we create an ECR for only the first (`0`) repository.

aws/ecr/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  description = "AWS region"
+}
+
+variable "zone_name" {
+  type        = "string"
+  description = "DNS zone name"
+}
+
+variable "repositories_names" {
+  type        = "list"
```
osterman: Seems like this should be a string or our ecr module should be made to support lists.

PR #21, opened 2018-07-03
Implement SES Email Forwarading

goruha:
## What
* Create email forwarder based on `ses` and `lambda`

## Why
* Allow to define and forward technical emails


aws/root-iam/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+## Spaces before and after `backend` required to select right word, because `backend` appears 3 times in main.tf
+sed -Ei 's/^(\s+role_arn\s+)/#\1/' main.tf
+
+init-terraform
+echo "yes" | terraform apply
```
osterman: Use the `terraform apply -auto-approve` instead

PR #29, opened 2018-07-20
Grant premissions for app secrets group

goruha:
## What
* Grant chamber permissions for `app` secrets group

## Why
* We use pattern `chamber exec kops app` so need access to both

### DIFF
--- a/aws/chamber/main.tf
+++ b/aws/chamber/main.tf
@@ -30,7 +30,7 @@ variable "account_id" {
 
 variable "parameter_groups" {
   type        = "list"
-  description = "Paramter groups names"
+  description = "Parameter group names"
   default     = ["kops", "app"]
 }
 
--- a/aws/chamber/main.tf
+++ b/aws/chamber/main.tf
@@ -28,6 +28,12 @@ variable "account_id" {
   description = "AWS account ID"
 }
 
+variable "parameter_groups" {
+  type        = "string"
+  description = "Comma separated name of paramter groups"
+  default     = "kops, app"
+}
+
 provider "aws" {
   assume_role {
     role_arn = "${var.aws_assume_role_arn}"
--- a/aws/chamber/main.tf
+++ b/aws/chamber/main.tf
@@ -29,9 +29,9 @@ variable "account_id" {
 }
 
 variable "parameter_groups" {
-  type        = "string"
-  description = "Comma separated name of paramter groups"
-  default     = "kops, app"
+  type        = "list"
+  description = "Paramter groups names"
+  default     = ["kops", "app"]
 }
 
 provider "aws" {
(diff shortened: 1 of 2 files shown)


aws/chamber/user.tf
```diff
@@ -7,7 +7,10 @@ module "chamber_user" {
   name          = "chamber"
   attributes    = ["codefresh"]
   kms_key_arn   = "${module.chamber_kms_key.key_arn}"
-  ssm_resources = ["${format("arn:aws:ssm:%s:%s:parameter/kops/*", var.region, var.account_id)}"]
+  ssm_resources = [
+    "${format("arn:aws:ssm:%s:%s:parameter/kops/*", var.region, var.account_id)}",
+    "${format("arn:aws:ssm:%s:%s:parameter/app/*", var.region, var.account_id)}"
```
osterman: app is more symbolic. I think each app should get it's own namespace. I think we should change this to a `formatlist` and introduce a variable for the chamber services. Then we can pass it `kops`, `foo` and `bar`. 

Also, Chamber with codefresh should be deprecated due to secrets getting exposed by helm values.

aws/chamber/main.tf
```diff
@@ -28,6 +28,12 @@ variable "account_id" {
   description = "AWS account ID"
 }
 
+variable "parameter_groups" {
+  type        = "string"
```
osterman: This should use the native type of `list`

osterman: You can use JSON notation in envs. e.g. `["one", "two"]`

goruha: then it would be hard to manage with ENV vars

aws/chamber/main.tf
```diff
@@ -28,6 +28,12 @@ variable "account_id" {
   description = "AWS account ID"
 }
 
+variable "parameter_groups" {
+  type        = "list"
+  description = "Paramter groups names"
```
aknysh: Parameter group names

aws/chamber/user.tf
```diff
@@ -1,13 +1,16 @@
...
 # https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-access.html
 module "chamber_user" {
-  source        = "git::https://github.com/cloudposse/terraform-aws-iam-chamber-user.git?ref=tags/0.1.4"
-  namespace     = "${var.namespace}"
-  stage         = "${var.stage}"
-  name          = "chamber"
-  attributes    = ["codefresh"]
-  kms_key_arn   = "${module.chamber_kms_key.key_arn}"
-  ssm_resources = ["${format("arn:aws:ssm:%s:%s:parameter/kops/*", var.region, var.account_id)}"]
+  source      = "git::https://github.com/cloudposse/terraform-aws-iam-chamber-user.git?ref=tags/0.1.4"
```
aknysh: Update to the latest version:
https://github.com/cloudposse/terraform-aws-iam-chamber-user.git?ref=tags/0.1.5

PR #34, opened 2018-08-02
Parameterize kops state backend

osterman:
## what
* Add variables (with defaults) for `ssh_public_key_path` and `name`

## why
* For a customer implementation we're writing public key to non-standard path


aws/kops/main.tf
```diff
@@ -38,7 +50,8 @@ module "kops_state_backend" {
   source           = "git::https://github.com/cloudposse/terraform-aws-kops-state-backend.git?ref=tags/0.1.3"
   namespace        = "${var.namespace}"
   stage            = "${var.stage}"
-  name             = "kops-state"
+  name             = "${var.name}"
+  attributes       = "state"
```
aknysh: should be a list:
attributes       = ["state"]

PR #35, opened 2018-08-21
[ecr] Fix Outputs

goruha:
## What
* Fix incorrect variable references to outputs

## Why
* Invalid


osterman:
Rename file to indicate relationship to Kops.

PR #36, opened 2018-08-21
Split backing services

goruha:
## What
* Split backing services to several directories 

## Why
* Not all projects needs `postgress` or `elasticache-redis`


osterman:
See: https://github.com/cloudposse/terraform-root-modules/pull/35#pullrequestreview-148153743

PR #45, opened 2018-10-24
[aws/users,aws/account-settings] new modules

osterman:
## what
* Add a `users` module
* Add a `account-settings module

## why
* Make it easy to use GitOps to provision users

### DIFF
--- a/aws/users/main.tf
+++ b/aws/users/main.tf
@@ -10,6 +10,16 @@ provider "aws" {
   }
 }
 
+data "terraform_remote_state" "account_settings" {
+  backend = "s3"
+
+  config {
+    bucket = "${var.namespace}-${var.stage}-terraform-state"
+    key    = "account-settings/terraform.tfstate"
+  }
+}
+
+
 data "terraform_remote_state" "root_iam" {
   backend = "s3"
 
@@ -26,6 +36,7 @@ output "account_alias" {
 }
 
 locals {
+  signin_url = "${data.terraform_remote_state.account_settings.signin_url}"
   admin_group = ["${data.terraform_remote_state.root_iam.admin_group}", "${data.terraform_remote_state.root_iam.readonly_group}"]
   readonly_group = ["${data.terraform_remote_state.root_iam.readonly_group}"]
 }
--- a/aws/users/main.tf
+++ b/aws/users/main.tf
@@ -19,7 +19,6 @@ data "terraform_remote_state" "account_settings" {
   }
 }
 
-
 data "terraform_remote_state" "root_iam" {
   backend = "s3"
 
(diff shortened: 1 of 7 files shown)


aws/users/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+  description = "Stage (e.g. `prod`, `dev`, `staging`)"
+}
+
+variable "name" {
+  type        = "string"
+  description = "Application or solution name (e.g. `app`)"
+  default     = "terraform"
+}
+
+variable "smtp_username" {
```
aknysh: looks like all vars below are not in use

aws/users/main.tf
```diff
@@ -0,0 +1,42 @@
...
+}
+
+data "aws_iam_account_alias" "default" {}
+
+output "account_alias" {
+  value = "${data.aws_iam_account_alias.default.account_alias}"
+}
+
+locals {
+  signin_url = "${data.terraform_remote_state.account_settings.signin_url}"
```
aknysh: `terraform fmt`

aknysh: locals not in use?

aws/users/variables.tf
osterman: See https://github.com/cloudposse/root.cloudposse.co/pull/24

aws/users/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  admin_group = ["${data.terraform_remote_state.root_iam.admin_group}", "${data.terraform_remote_state.root_iam.readonly_group}"]
```
aknysh: should it be just one admin group in the list?

aws/users/main.tf
aknysh: are we missing the actual `user` files?

aws/users/main.tf
osterman: i struggled with naming. my thinking is that admin has admin & readonly (why shouldn't they). readonly people only have readonly. i'm overloading `admin` which i don't like. but i don't know if it's better to add more terms. feels like it's implied that an admin would get readonly. thoughts?

aws/users/main.tf
osterman: See https://github.com/cloudposse/root.cloudposse.co/pull/24

aws/users/main.tf
```diff
@@ -0,0 +1,42 @@
...
+data "terraform_remote_state" "root_iam" {
+  backend = "s3"
+
+  config {
+    bucket = "${var.namespace}-${var.stage}-terraform-state"
+    key    = "root-iam/terraform.tfstate"
+  }
+}
+
+data "aws_iam_account_alias" "default" {}
```
aknysh: why use data provider and not take the alias from `"terraform_remote_state" "account_settings"`?

aws/users/variables.tf
aknysh: ok I see

aws/users/main.tf
aknysh: why not use the `admin` and `readonly` group names from `data.terraform_remote_state.root_iam`, and then in the account modules add the user to one of the group or to both?

```hcl
groups  = ["${local.admin_group}, ${local.readonly_group}"]
```

osterman: My bad. fixed.

PR #53, opened 2018-12-13
[aws/backing-services] Allow Kops Masters to Forward Logs to Elasticsearch

goruha:
## What
* Grant access to elastic search from masters

## Why
* To collect logs from master


aws/ecr/kops_ecr_user.tf
```diff
@@ -1,5 +1,5 @@
 module "kops_ecr_user" {
-  source    = "git::https://github.com/cloudposse/terraform-aws-iam-system-user.git?ref=tags/0.3.0"
+  source    = "git::https://github.com/cloudposse/terraform-aws-iam-system-user.git?ref=proxy-role-for-user"
```
osterman: Pin to release

aws/ecr/kops_ecr_app.tf
```diff
@@ -3,14 +3,14 @@ variable "kops_ecr_app_repository_name" {
 }
 
 module "kops_ecr_app" {
-  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.4"
+  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=use-roles-instead-of-users"
```
osterman: Pin to release

goruha:
```suggetion
  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.4"
```

goruha:
```suggestion
  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.4"
```

osterman: It's currently pinned to `use-roles-instead-of-users`

aws/ecr/kops_ecr_user.tf
goruha:
```suggestion
  source    = "git::https://github.com/cloudposse/terraform-aws-iam-system-user.git?ref=tags/0.3.0"
```

aws/backing-services/elasticsearch.tf
```diff
@@ -54,12 +54,19 @@ variable "ELASTICSEARCH_PERMITTED_NODES" {
...
     nodes   = ["${module.kops_metadata.nodes_role_arn}"]
     both    = ["${module.kops_metadata.masters_role_arn}", "${module.kops_metadata.nodes_role_arn}"]
     any     = ["*"]
   }
+
+  security_groups = {
+    masters = ["${module.kops_metadata.masters_security_group_id}"]
+    nodes   = ["${module.kops_metadata.nodes_security_group_id}"]
+    both    = ["${module.kops_metadata.masters_security_group_id}", "${module.kops_metadata.nodes_security_group_id}"]
+    any     = ["${module.kops_metadata.masters_security_group_id}", "${module.kops_metadata.nodes_security_group_id}"]
```
osterman: Why `both` and `any`?

goruha: `both` and `any` to be consistent with other terraform modules.
But it the case `both` and `any` have the same meaning because no way to grant access from any security group.

osterman:
@goruha rebase

goruha:
@osterman Done

PR #56, opened 2018-12-18
Use remote state for nameservers

osterman:
## what
* Use remote state and organizational account roles to obtain name servers for subaccount
* Define a local module to reduce repetition

## why
* *Radically* simplify the cold-start problem as it relates to DNS by eliminating the need to set environment variables for subaccounts
* Setting these environment variables in the `Dockerfile` was like playing the [Towers of Hanoi](https://en.wikipedia.org/wiki/Tower_of_Hanoi)

### DIFF
--- a/aws/root-dns/ns/variables.tf
+++ b/aws/root-dns/ns/variables.tf
@@ -1,0 +1,50 @@
+variable "namespace" {
+  type        = "string"
+  description = "Namespace (e.g. `eg` or `example`)"
+}
+
+variable "stage" {
+  type        = "string"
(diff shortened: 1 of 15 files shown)


aws/root-dns/ns/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+  description = "DNS zone to update"
+}
+
+variable "ttl" {
+  description = "Default TTL for the NS records"
+  default     = "30"
+}
+
+variable "key" {
+  default = "account-dns/terraform.tfstate"
```
aknysh: missing `description`

PR #59, opened 2018-12-19
[aws/ecr] Workaround for roles attachments limits

goruha:
## What
* Solve iam policy attachment limit

## Why
* Allow create multiple ECR


aws/ecr/kops_ecr_user.tf
```diff
@@ -9,6 +16,49 @@ module "kops_ecr_user" {
...
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:GetRepositoryPolicy",
+      "ecr:DescribeRepositories",
+      "ecr:ListImages",
+      "ecr:DescribeImages",
+      "ecr:BatchGetImage",
+    ]
+
+    resources = ["arn:aws:ecr::${data.aws_caller_identity.current.account_id}:repository/*"]
```
osterman: Make this a parameter that defaults to `*`

aws/ecr/kops_ecr_user.tf
```diff
@@ -9,6 +16,49 @@ module "kops_ecr_user" {
...
+  }
+}
+
+resource "aws_iam_policy" "default" {
+  name        = "${module.label.id}"
+  description = "Allow IAM Users to access ECR registries"
+  policy      = "${data.aws_iam_policy_document.default.json}"
+}
+
+resource "aws_iam_policy_attachment" "default" {
```
aknysh: better not to use `aws_iam_policy_attachment` at all, use `aws_iam_user_policy_attachment`.
see the WARNING here https://www.terraform.io/docs/providers/aws/r/iam_policy_attachment.html

goruha:
use this solution https://github.com/cloudposse/terraform-root-modules/pull/65

PR #64, opened 2018-12-21
Bootstrapping

osterman:
## what
* Add makefiles

## why
* Improve bootstrapping consistency


aws/root-iam/README.md
```diff
@@ -0,0 +1,5 @@
+# root-iam
+
+This module is responsible for settnig up the access groups in the root account.
```
aknysh:
```suggestion
This module is responsible for setting up the access groups in the root account. 
```

PR #65, opened 2018-12-21
[aws/ecr] Authorize a user to access multiple ECR repos

goruha:
## What
* Use new version of ECR `0.4.0`
* Deprecate `terraform-aws-kops-ecr`
* Grant permission to access ECR using ECR policy with a principal that has access to it. Basically, let ECR describe who can access it, rather than polices per ecr to user.

## Why
*  To solve IAM limit problem (more scalable strategy and probably the way we should have done it from the get go)
* To make it scalable

## Breaking changes
* Variable `roles` replaced with `principals_full_access` or `principals_readonly_access` and expects list or role\user arns as value
* User should have permissions
```
data "aws_iam_policy_document" "login" {
  statement {
    sid       = "ECRGetAuthorizationToken"
    effect    = "Allow"
    actions   = ["ecr:GetAuthorizationToken"]
    resources = ["*"]
  }
}
```
* We removed policies that provide access to the registry. (`policy_login_name`, `policy_login_arn`, `policy_read_name`, `policy_read_arn`, `policy_write_name`, `policy_write_arn`).
So you do not need to attach the policies to IAM role\user. Please provide IAM role\user arn as variable `principals_full_access` or `principals_readonly_access` depend on what type of access to you need.

Example:

```
module "kops_ecr" {
  source       = "git::https://github.com/cloudposse/terraform-aws-ecr.git?ref=tags/0.2.11"
  name         = "${var.name}"
  namespace    = "${var.namespace}"
  stage        = "${var.stage}"
  use_fullname = "${var.use_fullname}"

  roles = [
    "${module.kops_metadata.masters_role_name}",
    "${module.kops_metadata.nodes_role_name}",
  ]
}

resource "aws_iam_policy_attachment" "login" {
  count      = "${signum(length(var.users))}"
  name       = "${module.label.id}"
  users      = ["${var.users}"]
  policy_arn = "${module.kops_ecr.policy_login_arn}"
}

```

now should be 

```
module "kops_ecr" {
  source       = "git::https://github.com/cloudposse/terraform-aws-ecr.git?ref=tags/0.3.0"
  name         = "${var.name}"
  namespace    = "${var.namespace}"
  stage        = "${var.stage}"
  use_fullname = "${var.use_fullname}"

  principals_readonly_access = [
    "${module.kops_metadata.masters_role_arn}",
    "${module.kops_metadata.nodes_role_arn}",
  ]

  principals_full_access =  [
    "${var.users_arns}"
  ]
}

```


aws/ecr/kops_ecr_app.tf
```diff
@@ -3,14 +3,18 @@ variable "kops_ecr_app_repository_name" {
...
 module "kops_ecr_app" {
-  source       = "git::https://github.com/cloudposse/terraform-aws-kops-ecr.git?ref=tags/0.1.4"
-  namespace    = "${var.namespace}"
-  stage        = "${var.stage}"
-  name         = "${var.kops_ecr_app_repository_name}"
-  cluster_name = "${var.region}.${var.zone_name}"
-
-  users = [
-    "${module.kops_ecr_user.user_name}",
+  source    = "git::https://github.com/cloudposse/terraform-aws-ecr.git?ref=fix-iam-limit-solution-2"
```
osterman: Pin to release

osterman:
Can you update the description to better describe what this is doing differently

osterman:
@goruha let's address the comments and merge

PR #70, opened 2018-12-27
[aws/accounts] do not use ssm module

osterman:
## what
* Define `aws_ssm_parameter` parameters directly rather than rely on ssm module
* Use a submodule to reduce copy/pasta mistakes

## why
* `value of 'count' cannot be computed` errors
* Do not want to update `terraform-aws-ssm-parameter-store` to pass in count explicitly (no added benefit at this point)

## upgrading
* You'll need to migrate the state for accounts to the submodule

```
terraform state mv aws_organizations_account.audit  module.audit.aws_organizations_account.default
terraform state mv aws_organizations_account.staging  module.staging.aws_organizations_account.default
# ...
```

### DIFF
--- a/aws/accounts/testing.tf
+++ b/aws/accounts/testing.tf
@@ -1,53 +1,21 @@
-resource "aws_organizations_account" "testing" {
-  count                      = "${local.testing_count}"
-  name                       = "testing"
-  email                      = "${format(var.account_email, "testing")}"
-  iam_user_access_to_billing = "${var.account_iam_user_access_to_billing}"
-  role_name                  = "${var.account_role_name}"
-}
-
-locals {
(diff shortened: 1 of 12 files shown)


aws/accounts/testing.tf
```diff
@@ -1,44 +1,43 @@
...
-      overwrite   = "true"
-      description = "AWS Organization Account Access Role"
-    },
-  ]
+resource "aws_ssm_parameter" "testing_organization_account_access_role" {
+  count       = "${local.testing_count}"
+  name        = "/${var.namespace}/testing/organization_account_access_role"
+  description = "AWS Organization Account Access Role"
+  type        = "String"
+  value       = "${local.testing_account_id}"
```
aknysh:
```suggestion
  value       = "${local. testing_organization_account_access_role}"
```

PR #73, opened 2019-01-02
[kops] Write calculated settings to chamber

osterman:
## what
* Write calculated settings (subnets, s3 buckets, cluster names, etc) to SSM parameter store 
* Calculate availability zones (if none passed)
* Calculate subnet CIDRs

## why
* So we can read them using chamber
* Simplify the cold-start process by not requiring the user to copy/paste data
* Reduce error-prone subnet calculations (especially by those not used to defining subnets)

## references
* https://github.com/terraform-providers/terraform-provider-aws/issues/1082


osterman:
> why do we need to perform a merge on AZs? why not just use them if provided in an input list, otherwise use data lookup

a) using a ternary with lists is more complicated (terraform complains that you cannot use a ternary with lists. I am sure there's a workaround, but this current solution seems more elegant to me) 
b) documenting that the two variables are mutually exclusive with each other is more effort. no way to validate they are not both passed.
c) this allows a user to specify mandatory AZs and the rest are computed
d) user can specify total desired AZs

aws/kops/main.tf
```diff
@@ -85,11 +37,118 @@ module "kops_state_backend" {
...
+  source       = "subnets"
+  iprange      = "${module.private_subnets.cidrs[0]}"
+  newbits      = "${var.utility_subnets_newbits > 0 ? var.utility_subnets_newbits : local.availability_zone_count}"
+  netnum       = "${var.utility_subnets_netnum}"
+  subnet_count = "${local.availability_zone_count}"
+}
+
+locals {
+  private_subnets = "${join(",", slice(module.private_subnets.cidrs, 1, local.availability_zone_count+1))}"
+  utility_subnets = "${join(",", module.utility_subnets.cidrs)}}"
```
aknysh:
```suggestion
  utility_subnets = "${join(",", module.utility_subnets.cidrs)}"
```

aws/kops/main.tf
```diff
@@ -85,11 +37,118 @@ module "kops_state_backend" {
...
+      type        = "String"
+      overwrite   = "true"
+      description = "Kops cluster name"
+    },
+    {
+      name        = "${format(var.chamber_parameter_name, local.chamber_service, "kops_network_cidr")}"
+      value       = "${var.network_cidr}"
+      type        = "String"
+      overwrite   = "true"
+      description = "Kops private subnet CIDRs"
```
aknysh:
```suggestion
      description = "CIDR block of kops virtual network"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+}
+
+variable "availability_zones" {
+  type        = "list"
+  description = "List of availability zones in which to provision the cluster (should be an odd number to avoid split-brain)."
+  default     = []
+}
+
+variable "availability_zone_count" {
+  description = "Number of availability zones."
```
aknysh:
```suggestion
  description = "Number of availability zones to use (to provision the kops cluster in)"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+  default     = "$${name}.$${parent_zone_name}"
+}
+
+variable "network_cidr" {
+  description = "This is the CIDR block of your virtual network"
+  default     = "172.20.0.0/16"
+}
+
+variable "private_subnets_newbits" {
+  description = "This is the new mask for the subnet within the virtual network"
```
aknysh:
```suggestion
  description = "This is the new mask for the private subnet within the virtual network"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+  default     = "-1"
+}
+
+variable "private_subnets_netnum" {
+  description = "This is the zero-based index of the subnet when the network is masked with the newbit"
+  default     = "0"
+}
+
+variable "utility_subnets_newbits" {
+  description = "This is the new mask for the subnet within the virtual network."
```
aknysh:
```suggestion
  description = "This is the new mask for the utility subnet within the virtual network"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+  default     = "172.20.0.0/16"
+}
+
+variable "private_subnets_newbits" {
+  description = "This is the new mask for the subnet within the virtual network"
+  default     = "-1"
+}
+
+variable "private_subnets_netnum" {
+  description = "This is the zero-based index of the subnet when the network is masked with the newbit"
```
aknysh:
```suggestion
  description = "This is the zero-based index of the private subnet when the network is masked with the `newbit`"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+  default     = "0"
+}
+
+variable "utility_subnets_newbits" {
+  description = "This is the new mask for the subnet within the virtual network."
+  default     = "-1"
+}
+
+variable "utility_subnets_netnum" {
+  description = "This is the zero-based index of the subnet when the network is masked with the newbit"
```
aknysh:
```suggestion
  description = "This is the zero-based index of the utility subnet when the network is masked with the `newbit`"
```

aws/kops/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+  default     = "0"
+}
+
+variable "chamber_service" {
+  default = ""
```
aknysh:
```suggestion
  default     = ""
  description = "`chamber` service name. See [chamber usage](https://github.com/segmentio/chamber#usage) for more details"
```

osterman:
@aknysh I reapplied these changes. looks good.

osterman:
> is this to deploy kops in our own VPC in our subnets?

No. The intention is to minimize the “edit this here”, “rebuild”, “copy the outputs from there”, “map these settings to these other settings”.

so here are my problems
1) customer has 7 aws accounts and 4 kubernetes clusters. I don't want to calculate everything by hand.
2) accounts shouldn’t have overlapping CIDRs so we can support peering if needed
3) i suck at subnetting
4) so I want to make it easier to provision and document the process

---
there are pros/cons with this approach
1) with `chamber`, we cannot use `build-kops-manifest` as a `docker` build step. we can though export those settings to a `.envrc` (chamber supports that)
2) we can’t export to a `.envrc` b/c we have the `kiam` keys in chamber which are multi-line and chamber exports bad envrc :smiley:. the fix is to use `certmanager` to generate the self-signed certs for kiam and remove those settings from chamber
---
I think there may be a couple PRs over the coming weeks that will “clean up the UX” related to these changes. There are a lot of moving pieces / dirt needs to settle.

osterman:
This is the related documentation PR: https://github.com/cloudposse/geodesic/pull/339

osterman:
These changes should not introduce any backward incompatibilties

PR #101, opened 2019-01-25
[aws/codefresh-onprem] Add Enterprise backing services

joshmyers:
## what

PR to add CodeFresh Enterprise backing services to terraform-root-modules

## why

so we can provision these CodeFresh Enterprise backing services into the backing service VPC and subnets.

Note that I have had to add some new outputs to the `backing-services` module so that I can import these as remote state to pass into the `terraform-aws-codefresh-backing-services` module

### DIFF
--- a/aws/codefresh-onprem/main.tf
+++ b/aws/codefresh-onprem/main.tf
@@ -1,0 +1,54 @@
+terraform {
+  required_version = ">= 0.11.2"
+
+  backend "s3" {}
+}
+
+provider "aws" {
+  assume_role {
+    role_arn = "${var.aws_assume_role_arn}"
+  }
+}
+
+variable "aws_assume_role_arn" {
+  type = "string"
+}
+
+variable "namespace" {
+  type        = "string"
+  description = "Namespace (e.g. `eg` or `cp`)"
+}
+
+variable "stage" {
+  type        = "string"
+  description = "Stage (e.g. `prod`, `dev`, `staging`)"
+}
+
+variable "region" {
+  type        = "string"
+  description = "AWS region"
+}
+
+variable "zone_name" {
+  type        = "string"
+  description = "DNS zone name"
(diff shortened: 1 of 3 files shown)


aws/codefresh-onprem/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+module "codefresh_enterprise_backing_services" {
+  source          = "git::https://github.com/cloudposse/terraform-aws-codefresh-backing-services.git?ref=initial_implementation"
+  namespace       = "${var.namespace}"
+  stage           = "${var.stage}"
+  vpc_id          = "${data.terraform_remote_state.backing_services.vpc_id}"
+  subnet_ids      = ["${data.terraform_remote_state.backing_services.private_subnet_ids}"]
+  security_groups = ["${concat(module.kops_metadata.masters_security_group_id,module.kops_metadata.nodes_security_group_id)}"]
```
goruha:
```suggestion
  security_groups = ["${concat(module.kops_metadata.nodes_security_group_id)}"]
```

aws/codefresh-onprem/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  type        = "string"
+  description = "AWS region"
+}
+
+variable "zone_name" {
+  type        = "string"
+  description = "DNS zone name"
+}
+
+data "terraform_remote_state" "backing_services" {
```
goruha: Not sure if we need this.
@osterman what do you think?

joshmyers: @goruha how else can we fetch the backing services VPC and subnets so that we can inject the CodeFresh Enterprise stuff into the same VPC/subnets?

osterman: Yes, I like this =)

PR #117, opened 2019-02-14
[aws/chamber] add 'enable' variable

tamsky:
## What's changing

Adds `enabled` param when calling module `cloudposse/terraform-aws-iam-chamber-user`.
Module already supports it.
Maintains historical users: default=`true`

## Why

We have an environment where we'd like to set this to disabled.

### DIFF
--- a/aws/chamber/user.tf
+++ b/aws/chamber/user.tf
@@ -10,6 +10,7 @@ module "chamber_user" {
   name        = "chamber"
   attributes  = ["codefresh"]
   kms_key_arn = "${module.chamber_kms_key.key_arn}"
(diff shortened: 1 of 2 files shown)


tamsky:
This all seems very similar to https://github.com/cloudposse/terraform-root-modules/pull/116
=)

osterman:
@tamsky sorry for that! my fault for not managing correctly

osterman:
see @goruha 's PR though as it uses more precise variable to indicate what's being enabled/disabled, which is desirable in this case, sinc we're not enabling and disabling the entire module.

PR #121, opened 2019-02-23
[aws/account-settings] add var to module: minimum_password_length

tamsky:
## What's changing

Adding a new variable `minimum_password_length` to the `aws/account-settings` module.

## Why

Account settings should write and output the account-level minimum password lengths so that
our https://github.com/cloudposse/terraform-aws-iam-user module can use the output value when createing users, so valid password lengths can be used.

### DIFF
--- a/aws/account-settings/variables.tf
+++ b/aws/account-settings/variables.tf
@@ -1,6 +1,7 @@
 variable "minimum_password_length" {
   type        = "string"
   description = "Minimum number of characters allowed in an IAM user password.  Integer between 6 and 128, per https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_account-policy.html"
+  default     = "8"  ## Same default as https://github.com/cloudposse/terraform-aws-iam-account-settings
 }
 
 variable "aws_assume_role_arn" {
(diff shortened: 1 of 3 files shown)


aws/account-settings/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "minimum_password_length" {
+  type        = "string"
+  description = "Minimum number of characters allowed in an IAM user password.  Integer between 6 and 128, per https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_account-policy.html"
```
tamsky: @osterman :: does this new variable need a `default` value?

I'm thinking that committing it "as-is" will break upstream users by requiring a new variable and value declaration.

osterman: I would prefer to set a sane default

tamsky: module we call currently has a default of `8`:
https://github.com/cloudposse/terraform-aws-iam-account-settings/pull/1/files#diff-8e2475e53e84b1e038e9f5d670c3faefR12

tamsky: Fixed in https://github.com/cloudposse/terraform-root-modules/pull/121/commits/2292b1c1cab61817dfae8a2f502aaa9c483e5516

PR #125, opened 2019-03-06
[aws/kops-iam-users] Include terraform Makefile and example vars

Nuru:
## what
Add Makefile to aws/kops-iam-users/

## why
Required by Atlantis


Nuru:
data and corp are not good examples

PR #135, opened 2019-03-18
[kops-aws-platform] Disable kops-iam by default

Nuru:
## what
Disable kops iam integration by default

## why
Most users do not want kops iam integration, but prior to this change, it was forced on them.

### DIFF
--- a/aws/kops-aws-platform/iam-authenticator.tf
+++ b/aws/kops-aws-platform/iam-authenticator.tf
@@ -1,6 +1,13 @@
+variable "kops_iam_enabled" {
+  type        = "string"
+  description = "Set to true to allow the module to create EFS resources"
+  default     = "false"
+}
+
 variable "cluster_id" {
   type        = "string"
   description = "A unique-per-cluster identifier to prevent replay attacks. Good choices are a random token or a domain name that will be unique to your cluster"
+  default     = "random"
 }
 
 variable "kube_config_path" {
@@ -34,6 +41,8 @@ variable "readonly_k8s_groups" {
 }
 
 resource "kubernetes_cluster_role_binding" "view" {
+  count = "${var.kops_iam_enabled == "true" ? 1 : 0}"
+
   metadata {
     name = "view-binding"
   }
@@ -64,7 +73,7 @@ module "kops_admin_label" {
   attributes = ["admin"]
   delimiter  = "${var.delimiter}"
   tags       = "${var.tags}"
-  enabled    = "true"
+  enabled    = "${var.kops_iam_enabled}"
 }
 
 module "kops_readonly_label" {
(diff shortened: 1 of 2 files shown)


aws/kops-aws-platform/iam-authenticator.tf
```diff
@@ -1,6 +1,13 @@
+variable "kops_iam_enabled" {
+  type        = "string"
+  description = "Set to true to allow the module to create EFS resources"
```
aknysh: update description (remove EFS)

aws/kops-aws-platform/iam-authenticator.tf
```diff
@@ -115,19 +129,21 @@ data "aws_iam_policy_document" "admin" {
...
 }
 
 module "iam_authenticator_config" {
-  source                = "git::https://github.com/cloudposse/terraform-aws-kops-iam-authenticator-config.git?ref=tags/0.1.1"
+  enabled               = "${var.kops_iam_enabled}"
+  source                = "git::https://github.com/cloudposse/terraform-aws-kops-iam-authenticator-config.git?ref=tags/0.2.0"
   cluster_id            = "${var.cluster_id}"
   kube_config_path      = "${var.kube_config_path}"
-  admin_iam_role_arn    = "${aws_iam_role.admin.arn}"
+  admin_iam_role_arn    = "${element(concat(aws_iam_role.admin.*.arn, list("")), 0)}"
```
aknysh:
```suggestion
  admin_iam_role_arn    = "${join("", aws_iam_role.admin.*.arn)}"
```

aws/kops-aws-platform/iam-authenticator.tf
```diff
@@ -115,19 +129,21 @@ data "aws_iam_policy_document" "admin" {
...
   admin_k8s_username    = "${var.admin_k8s_username}"
   admin_k8s_groups      = "${var.admin_k8s_groups}"
-  readonly_iam_role_arn = "${aws_iam_role.readonly.arn}"
+  readonly_iam_role_arn = "${element(concat(aws_iam_role.readonly.*.arn, list("")), 0)}"
```
aknysh:
```suggestion
  readonly_iam_role_arn = "${join("", aws_iam_role.readonly.*.arn)}"
```

aws/kops-aws-platform/iam-authenticator.tf
Nuru: This is the [canonical idiom from the Terraform docs](https://www.terraform.io/upgrade-guides/0-11.html#referencing-attributes-from-resources-with-count-0), and in public code I like to follow conventions when it is reasonable. I agree that `join` is shorter and cleaner, but is there a functional difference?

aknysh: o functional difference (we just use it everywhere in our code)

PR #145, opened 2019-04-04
[aws/tfstate-backend] update to version 0.7.0

Nuru:
## what
[aws/tfstate-backend] update from version 0.1.1 to version 0.7.0

## why
Enforce server-side encryption and pick up other new features

### DIFF
--- a/aws/tfstate-backend/main.tf
+++ b/aws/tfstate-backend/main.tf
@@ -1,9 +1,7 @@
 terraform {
   required_version = ">= 0.11.2"
 
-  backend "s3" {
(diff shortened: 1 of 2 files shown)


aws/tfstate-backend/main.tf
```diff
@@ -1,6 +1,9 @@
 terraform {
   required_version = ">= 0.11.2"
-  backend          "s3"             {}
+
+  backend "s3" {
```
osterman: Guessing we are going to keep these globally?

PR #155, opened 2019-04-11
Add [kops-legacy-account-vpc-peering]

aknysh:
## what
* Add `kops-legacy-account-vpc-peering` project

## why
* Provision VPC peering connection between a `kops` VPC and a legacy VPC in a different AWS account (possibly provisioned manually without using Terraform or other automation tools)
* Specify what parameters are needed from the legacy AWS account to be able to provision a VPC peering connection
* Show the IAM role, policies and permissions required in the legacy account

### DIFF
--- a/aws/kops-legacy-account-vpc-peering/main.tf
+++ b/aws/kops-legacy-account-vpc-peering/main.tf
@@ -1,0 +1,40 @@
+terraform {
+  required_version = ">= 0.11.2"
+
+  backend "s3" {}
+}
+
+provider "aws" {
+  assume_role {
+    role_arn = "${var.aws_assume_role_arn}"
+  }
+}
+
+# Lookup VPC of the kops cluster
+module "kops_metadata" {
+  source   = "git::https://github.com/cloudposse/terraform-aws-kops-metadata.git?ref=tags/0.2.1"
+  dns_zone = "${var.region}.${var.zone_name}"
+  enabled  = "${var.enabled}"
+}
+
+module "kops_legacy_account_vpc_peering" {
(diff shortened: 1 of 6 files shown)


aws/kops-legacy-account-vpc-peering/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  enabled = "${var.enabled}"
+
+  namespace = "${var.namespace}"
+  stage     = "${var.stage}"
+  name      = "${var.name}"
+
+  auto_accept = true
+
+  # Requester
+  requester_vpc_id              = "${module.kops_metadata.vpc_id}"
```
Nuru: I don't think this actually works. It looks to me like `terraform-aws-kops-metadata` is going to return all the VPCs and you are just getting lucky that there is only one to return.

aknysh: We use it everywhere.

https://github.com/cloudposse/terraform-root-modules/blob/master/aws/backing-services/elasticsearch.tf#L58
https://github.com/cloudposse/terraform-root-modules/blob/master/aws/backing-services/aurora-postgres.tf#L92

It returns only one VPC.

If we are changing the way we name and tag `kops` resources, we need to discuss that since it could introduce breaking changes.

Nuru: `kops`/`kubernetes` 1.12 does not tag the VPC when it does not create the VPC and provides no consistent way to determine the VPC it is using. Probably the right place to address this in the long run is `terraform-aws-kops-metadata` but for now let's at least pass through [`vpc_tag` and `vpc_tag_vaules`](https://github.com/cloudposse/terraform-aws-kops-metadata/blob/cea5cba634b2659aefb6a2e6821161d7081baa1d/variables.tf#L29-L37)

PR #164, opened 2019-04-18
[backing services] Fix CIDR range because of new availability zone

goruha:
## What
* Added `vpc_max_subnets_count`

## Why
* Allow specifying the count of ranges CIDR is split to

### DIFF
--- a/aws/backing-services/vpc.tf
+++ b/aws/backing-services/vpc.tf
@@ -10,6 +10,12 @@ variable "vpc_nat_gateway_enabled" {
   default = "true"
 }
 
+variable "vpc_max_subnet_count" {
+  default     = 0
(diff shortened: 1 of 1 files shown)


aws/backing-services/vpc.tf
```diff
@@ -10,6 +10,11 @@ variable "vpc_nat_gateway_enabled" {
   default = "true"
 }
 
+variable "vpc_max_subnet_count" {
+  default     = 0
+  description = "Sets the maximum amount of subnets to deploy.  0 will deploy a subnet for every availablility zone within the region"
```
aknysh:
```suggestion
  description = "The maximum count of subnets to provision. 0 will provision a subnet for each availability zone within the region"
```

PR #176, opened 2019-05-24
Add `kops-aurora-postgres` module

aknysh:
## what
* Add `kops-aurora-postgres` module

## why
* To provision Aurora Postgres cluster in an existing `kops` VPC
* It's different from `backing-services/aurora-postgres.tf`, which provisions Aurora Postgres cluster in a backing services VPC which is separate from `kops` VPC

### DIFF
--- a/aws/kops-aurora-postgres/variables.tf
+++ b/aws/kops-aurora-postgres/variables.tf
@@ -1,0 +1,119 @@
+variable "aws_assume_role_arn" {
+  type = "string"
+}
+
+variable "namespace" {
+  type        = "string"
+  description = "Namespace (e.g. `eg` or `cp`)"
+}
+
+variable "stage" {
+  type        = "string"
+  description = "Stage (e.g. `prod`, `dev`, `staging`)"
+}
+
+variable "zone_name" {
+  type        = "string"
+  default     = ""
+  description = "Domain name of DNS zone in which to add subdomain records, optional, defaults to cluster_name"
+}
+
+variable "kops_cluster_name" {
+  type        = "string"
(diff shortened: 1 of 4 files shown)


Nuru:
Can we name this `aurora-postgres`, since VPC ID can be configured?

aws/kops-aurora-postgres/variables.tf
```diff
@@ -0,0 +1,119 @@
...
+
+variable "postgres_cluster_dns_name" {
+  type        = "string"
+  description = "Name of the cluster CNAME record to create in the parent DNS zone specified by `zone_name`. If left empty, the name will be auto-asigned using the format `master.var.postgres_name`"
+  default     = ""
+}
+
+variable "postgres_reader_dns_name" {
+  type        = "string"
+  description = "Name of the reader endpoint CNAME record to create in the parent DNS zone specified by `zone_name`. If left empty, the name will be auto-asigned using the format `replicas.var.postgres_name`"
```
Nuru: Why is this called "reader" instead of "replica"?

aknysh: https://www.terraform.io/docs/providers/aws/r/rds_cluster.html#reader_endpoint

aknysh:
VPC ID can be configured, but there are too many logical paths to make the module generic.
E.g. 

1) if you provide a VPC ID, do we need to use it for the cluster, or do we need to lookup the `kops` VPC by the provided ID

2) If you provide a VPC ID, do you need to provide also the subnets and SGs?
If you want to use existing subnets and SGs, then you do.
If you want to lookup the `kops` VPC, then you don't

Making the module generic will require a lot of logic to be correct

PR #179, opened 2019-06-04
[README] Update documentation on "usage"

Nuru:
## what
Update README documentation on "usage"
## why
Reflect revised practice of dynamically updating modules at run time rather than statically including them at build time

### DIFF
--- a/README.yaml
+++ b/README.yaml
@@ -121,18 +121,28 @@ introduction: |-
 
 # How to use this project
 usage: |-
-  Use the `terraform-root-modules` Docker image as the base image in the application `Dockerfile`, and copy the modules from `/aws` folder into `/conf` folder.
+  With the CloudPosse [reference architecture](https://github.com/cloudposse/reference-architectures), each environment (e.g. staging and prod) has its own, customized
+  version of [Geodesic](https://github.com/cloudposse/geodesic). The process for using one of these Terraform root modules
+  is one of further customizing the environment's Geodesic as follows:
+  1. Create a directory under `/conf` to contain the files related to the root module. Normally the directory has the same name
+  as the root module.
+  1. Copy the `.envrc`, `Makefile.tasks`, and `terraform.envrc` files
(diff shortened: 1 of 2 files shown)


README.yaml
```diff
@@ -121,18 +121,28 @@ introduction: |-
...
-
-  WORKDIR /conf/
-  ```
+  With the CloudPosse [reference architecture](https://github.com/cloudposse/reference-architectures), each environment (e.g. staging and prod) has its own, customized
+  version of [Geodesic](https://github.com/cloudposse/geodesic). The process for using one of these Terraform root modules
+  is one of further customizing the environment's Geodesic as follows:
+  1. Create a directory under `/conf` to contain the files related to the root module. Normally the directory has the same name
+  as the root module.
+  1. Copy the `.envrc`, `Makefile.tasks`, and `terraform.envrc` files
+  from [`cloudposse/reference-architectures/templates/conf/kops`](https://github.com/cloudposse/reference-architectures/tree/0.13.0/templates/conf/kops)
```
aknysh:
```suggestion
  from [`cloudposse/reference-architectures/templates/conf/kops`](https://github.com/cloudposse/reference-architectures/tree/master/templates/conf/kops)
```

better to use `master`

README.yaml
```diff
@@ -121,18 +121,28 @@ introduction: |-
...
+  into the new directory. The `.envrc` and `Makefile.tasks` files are boilerplate and do not need to be changed.
+  1. Edit the `terraform.envrc` file so that the `TF_CLI_INIT_FROM_MODULE` points to the desired Terraform root module.
+  Be sure to pin it to a specific version. Add additional environment variable settings if desired.
+  1. Add a `terraform.tfvars` file that sets the vaules
```
aknysh:
```suggestion
  1. Add a `terraform.tfvars` file that sets the values
```

README.yaml
Nuru: I thought we did not want to reference master. @osterman ???

PR #181, opened 2019-06-25
[vpc] Allow for deploying the VPC into fewer than all the availability zones

Nuru:
## what
Allow for deploying the VPC into fewer than all the availability zones
## why
We typically only want to deploy into 3 availability zones, but prior to this PR the VPC deployed into every available availability zone, and, for example, `us-east-1` has 6 availability zones

### DIFF
--- a/aws/vpc/main.tf
+++ b/aws/vpc/main.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 0.11.2"
+  required_version = "~> 0.11.2"
 
   backend "s3" {}
 }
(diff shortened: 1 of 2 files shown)


aws/vpc/main.tf
```diff
@@ -91,6 +101,15 @@ resource "aws_ssm_parameter" "nat_gateways" {
   overwrite   = "true"
 }
 
+resource "aws_ssm_parameter" "nat_instances" {
```
aknysh:
```suggestion
resource "aws_ssm_parameter" "nat_instance_ids" {
```

aws/vpc/main.tf
```diff
@@ -91,6 +101,15 @@ resource "aws_ssm_parameter" "nat_gateways" {
...
+  count       = "${var.vpc_nat_instance_enabled == "true" ? 1 : 0}"
+  name        = "${format(var.chamber_parameter_name, local.chamber_service, module.parameter_prefix.id, "nat_instances")}"
+  value       = "${join(",", module.subnets.nat_instance_ids)}"
+  description = "VPC private NAT instances"
```
aknysh:
```suggestion
  description = "NAT instance IDs"
```

PR #197, opened 2019-08-08
Add grafana

Nuru:
## what

Provision database for Grafana

## why

Grafana otherwise loses all customization when the pod is destroyed

### DIFF
--- a/aws/grafana-backing-services/Makefile
+++ b/aws/grafana-backing-services/Makefile
@@ -1,0 +1,15 @@
+## Initialize terraform remote state
+init:
+	[ -f .terraform/terraform.tfstate ] || terraform $@
+
+## Clean up the project
+clean:
+	rm -rf .terraform *.tfstate* $$TF_MODULE_CACHE
+
+## Pass arguments through to terraform which require remote state
+apply console destroy graph plan output providers show: init
+	terraform $@
+
+## Pass arguments through to terraform which do not require remote state
+get fmt validate version:
+	terraform $@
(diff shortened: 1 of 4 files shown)


aws/grafana-backing-services/Makefile
```diff
@@ -0,0 +1,15 @@
+## Initialize terraform remote state
+init:
+	[ -f .terraform/terraform.tfstate ] || terraform $@
+
+## Clean up the project
+clean:
+	rm -rf .terraform *.tfstate* $$TF_MODULE_CACHE
```
aknysh:
```suggestion
	rm -rf .terraform *.tfstate* "${TF_MODULE_CACHE:-.terraform}"
```

aws/grafana-backing-services/aurora-mysql.tf
```diff
@@ -0,0 +1,254 @@
+# https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html
+
+variable "mysql_name" {
+  type        = "string"
```
aknysh:
```suggestion
  type        = string
```

aws/grafana-backing-services/aurora-mysql.tf
```diff
@@ -0,0 +1,254 @@
...
+}
+
+variable "mysql_cluster_enabled" {
+  type        = "string"
+  default     = "false"
+  description = "Set to false to prevent the module from creating any resources"
+}
+
+variable "mysql_cluster_publicly_accessible" {
+  default     = "false"
```
aknysh:
```suggestion
  type        = bool
  default     = false
```

aws/grafana-backing-services/Makefile
Nuru: I don't know why I proposed "${TF_MODULE_CACHE:-.terraform}" and I don't think it works anyway. `$$TF_MODULE_CACHE` works and is nicer because if `TF_MODULE_CACHE` is not set, then it simply has no effect.

PR #201, opened 2019-08-21
account added to root-dns 

maximmi:
## what
* account added

## why
* to allow apply domains for existing stages


aws/root-dns/parent-qa-ns.tf
```diff
@@ -0,0 +1,12 @@
+module "qa" {
+  source           = "ns"
+  accounts_enabled = "${var.accounts_enabled}"
+  namespace        = "${var.namespace}"
+  stage            = "qa"
+  zone_id          = "${aws_route53_zone.parent_dns_zone.zone_id}"
+  account          = "staging"
```
osterman:
```suggestion
  account          = "staging"
  key                  = "qa-dns/terraform.tfstate"
```

PR #208, opened 2019-09-15
[kops-aws-platofrm] Store EFS info in SSM for use by efs-provisioner

Nuru:
## what
[kops-aws-platofrm] Store EFS info in SSM for use by efs-provisioner

## why
Avoid the need to copy and paste output of Terraform into input of Helmfile

### DIFF
--- a/aws/kops-aws-platform/efs-provisioner.tf
+++ b/aws/kops-aws-platform/efs-provisioner.tf
@@ -24,7 +24,7 @@ locals {
 }
 
 module "kops_efs_provisioner" {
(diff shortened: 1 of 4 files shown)


aws/kops-aws-platform/efs-provisioner.tf
```diff
@@ -39,6 +39,24 @@ module "kops_efs_provisioner" {
   }
 }
 
+resource "aws_ssm_parameter" "kops_efs_provisioner_role_name" {
+  count       = "${var.efs_enabled == "true" ? 1 : 0}"
+  name        = "/kops/kops_efs_provisioner_role_name"
```
aknysh: maybe parameterize `/kops` since we already discussed to change it to `global`

PR #227, opened 2020-02-26
Avoid all kibanas to have the same subdomain name

asehra:


goruha:
@asehra could you make it back compatible ?

Probably add a variable

PR #235, opened 2020-04-28
Add domain_name to outputs

jhenderson:
## what
Add `domain_name` to outputs.

## why
* We are configuring monitoring alerts in Elasticsearch that require the `domain_name`.


goruha:
@jhenderson you need to update version for elasticsearch module

goruha:
also new version is for TF 0.12, but this module use 0.11 so they are not compatible

jhenderson:
`elasticsearch_domain_name` already added as part of `0.124.0`

PR #250, opened 2020-06-15
module to create ECR repo for every github repo for organization

maximmi:
## what
* module to create `ECR` repo for every `github` repo for given organization

## why
* to automate creation of ECR repos

### DIFF
--- a/aws/github-org-ecr/main.tf
+++ b/aws/github-org-ecr/main.tf
@@ -33,3 +33,11 @@ module "ecr" {
   scan_images_on_push        = var.scan_images_on_push
   image_tag_mutability       = var.image_tag_mutability
 }
+
+output "repository_url_map" {
+  value = module.ecr.repository_url_map
+}
+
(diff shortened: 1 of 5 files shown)


aws/github-org-ecr/main.tf
```diff
@@ -0,0 +1,35 @@
...
+}
+
+provider "aws" {
+  assume_role {
+    role_arn = var.aws_assume_role_arn
+  }
+}
+
+provider "github" {
+  token        = var.github_token
```
aknysh, line 12: @maximmi 
Is that module to create ECR for all repos, including private, or just public repos?
If public, then we don't need to provide GH token, we can implement what was done here https://github.com/cloudposse/terraform-github-repository-webhooks/pull/18/files#diff-7a370d8342e7203b805911c92454f0f4R4

osterman: it's predominantly for *private* images ;)

PR #297, opened 2021-02-03
feat: adds component documentation + pre-commit + rebuild-docs target 

Gowiem:
## what
* Updates pre-commit to include terraform_docs + bumps to 0.45.0
* Updates pre-commit workflow for terraform_docs checking
* Adds information on installing / using pre-commit
* Adds rebuild-docs target
* Adds README for all components
* Updates root README with components table

## why
* Supports consumers by providing them additional documentation
* pre-commit checks ensure that documentation is kept up to date


modules/opsgenie/README.md
```diff
@@ -1,338 +1,86 @@
...
-      delay: 10
-      recipients:
-        - type: team
-          team_name: devops
-    repeat:
-      wait_interval: 10
-      count: 2
-      reset_recipient_states: false
-      close_alert_after_all: false
+TODO: Do we have a full example YAML snippet somewhere?
```
Gowiem, line 12: @osterman @aknysh Do either of you have a good example stack snippet for this module from another project that you could share? I figured a full example from a client project that has been made generic would be best here since it seems like it's highly customizable.

osterman: Yep! https://github.com/cloudposse/terraform-opsgenie-incident-management/tree/master/examples/config/resources

osterman (approved):

modules/opsgenie/README.md
Gowiem: @osterman updated 👍

osterman (approved):

PR #300, opened 2021-04-05
feat: updates EKS component to latest

Gowiem:
## what
* Updates the EKS component to the latest from downstream
* Updates pre-commit-terraform version to utilize latest terraform-docs version / README updates

## why
* We're going to be utilizing the EKS component in our upcoming tutorial so we wanted to be sure to be using the latest
* Bunch of terraform-docs updates recently, so utilizing the latest ensures we're keeping up-to-date with the improved README format. 

## references
* This broke the pre-commit GHA workflow, so I've separated the tf-docs version upgrade out into #301 which should be merged prior to this.

### DIFF
--- a/modules/eks/default.auto.tfvars
+++ b/modules/eks/default.auto.tfvars
@@ -14,7 +14,7 @@ cluster_endpoint_private_access = true
 
 cluster_endpoint_public_access = true
 
-cluster_kubernetes_version = "1.17"
+cluster_kubernetes_version = "1.18"
 
 enabled_cluster_log_types = ["api", "audit", "authenticator", "controllerManager", "scheduler"]
 
--- a/modules/eks/default.auto.tfvars
+++ b/modules/eks/default.auto.tfvars
@@ -4,6 +4,8 @@ enabled = true
 
 name = "eks"
 
+availability_zone_abbreviation_type = "short"
+
 allowed_security_groups = []
 
 allowed_cidr_blocks = []
@@ -14,7 +16,7 @@ cluster_endpoint_private_access = true
 
 cluster_endpoint_public_access = true
 
-cluster_kubernetes_version = "1.18"
+cluster_kubernetes_version = "1.19"
 
 enabled_cluster_log_types = ["api", "audit", "authenticator", "controllerManager", "scheduler"]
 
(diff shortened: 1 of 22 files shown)


modules/eks/default.auto.tfvars
```diff
@@ -14,7 +14,7 @@ cluster_endpoint_private_access = true
 
 cluster_endpoint_public_access = true
 
-cluster_kubernetes_version = "1.17"
+cluster_kubernetes_version = "1.18"
```
Gowiem, line 17: Note, bumping to version `1.18` since figured that is preferable. We can also update to `1.19` as well since that is out.

modules/eks/main.tf
```diff
@@ -1,8 +1,8 @@
 locals {
-  primary_role_map   = data.terraform_remote_state.primary_roles.outputs.role_name_role_arn_map
-  delegated_role_map = data.terraform_remote_state.delegated_roles.outputs.role_name_role_arn_map
-  eks_outputs        = data.terraform_remote_state.eks.outputs
-  vpc_outputs        = data.terraform_remote_state.vpc.outputs
+  primary_role_map   = module.primary_roles.outputs.role_name_role_arn_map
```
Gowiem, line 2: @osterman Will have to see how this component dependency shakes out in regards to our single AWS Account tutorial idea.

modules/eks/providers.tf
```diff
@@ -8,14 +8,29 @@ provider "aws" {
   }
 }
 
+provider "aws" {
```
Gowiem, line 11: @osterman you and I talked about this lightly but we didn't come up with an action item I don't believe... it seems funky that we're requiring spotinst for our EKS component. Do we potentially want to pull this out to its own component like `eks-with-spotinst` or a fully separate `spotinst-eks-node-group` component? cc @aknysh

Nuru (changes requested):
This is not close to the current eks component. Please upgrade to current 
- context.tf 0.25.0
- eks-cluster 0.43.2
- eks-node-group 0.26.2

Gowiem:
This is ancient as we never got around to getting it review and it looks like @nitrocode is working on the continuation of this work in #373. I'll let him take it from here 👍

PR #302, opened 2021-04-06
feat: updates VPC component to latest

Gowiem:
## what
* Updates the VPC component to the latest from downstream

## why
* We're going to be utilizing the VPC component in our upcoming tutorial so we wanted to be sure to be using the latest

## references
* #301 needs to be merged prior to this PR to fix the build

### DIFF
--- a/modules/vpc/main.tf
+++ b/modules/vpc/main.tf
@@ -14,7 +14,8 @@ locals {
 }
 
 module "vpc" {
-  source = "git::https://github.com/cloudposse/terraform-aws-vpc.git?ref=tags/0.18.0"
+  source  = "cloudposse/vpc/aws"
+  version = "0.18.0"
 
   tags       = local.tags
   cidr_block = var.cidr_block
@@ -22,6 +23,7 @@ module "vpc" {
   context = module.this.context
 }
 
+# required tags to make ALB ingress work https://docs.aws.amazon.com/eks/latest/userguide/alb-ingress.html
 # https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html
 locals {
   public_subnets_additional_tags = {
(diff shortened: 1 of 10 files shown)


modules/vpc/vpc-flow-logs.tf
```diff
@@ -0,0 +1,9 @@
+# https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/flow_log
+
+resource "aws_flow_log" "default" {
+  count                = module.this.enabled && var.vpc_flow_logs_enabled ? 1 : 0
+  log_destination      = var.vpc_flow_logs_destination
```
Gowiem, line 5: @aknysh @osterman I removed the usage of the `remote-state` module / file in this vpc component as it was only grabbing the `vpc-flow-logs-bucket` ARN from remote-state and it seemed to me to overly complicate things. I added the `vpc_flow_logs_destination` variable instead which means ya'll would need to add that as a var value in stacks. This is particularly simpler in the case where you do not want to enable VPC flow logs and / or create that flow logs bucket, which unfortunately is not supported because the `yaml-stack-config` module's `remote-state` module will err regardless of enabled or disabled. 

Anyway, let me know your thoughts on this one.

modules/vpc/main.tf
```diff
@@ -2,7 +2,8 @@ locals {
   # The usage of the specific kubernetes.io/cluster/* resource tags below are required
   # for EKS and Kubernetes to discover and manage networking resources
   # https://www.terraform.io/docs/providers/aws/guides/eks-getting-started.html#base-vpc-networking
-  tags = map(format("kubernetes.io/cluster/%s-%s-%s-eks-cluster", module.this.namespace, module.this.environment, module.this.stage), "shared")
+  cluster_tag_key = format("kubernetes.io/cluster/%s-%s-%s-eks-cluster", module.this.namespace, module.this.environment, module.this.stage)
+  tags            = tomap({ (local.cluster_tag_key) = "shared" })
```
Gowiem, line 6: 0.15.0+ required change.

PR #306, opened 2021-04-15
feat: upstreams datadog-integration and data-monitor components

Gowiem:
## what
* Upstreams datadog-integration and data-monitor components

## why


Gowiem:
Old as dirt!

PR #311, opened 2021-04-15
feat: upstreams eks-iam component

Gowiem:
## what
* Upstreams eks-iam component

## why


Gowiem:
Old as dirt!

PR #312, opened 2021-04-15
feat: upstreams cloudtrail and cloudtrail-bucket components

Gowiem:
## what
* Upstreams cloudtrail and cloudtrail-bucket components

## why


Gowiem:
Old as dirt!

PR #313, opened 2021-04-15
feat: adds aurora-postgres components

Gowiem:
## what
* Adds aurora-postgres components

## why


Gowiem:
@osterman @Nuru @aknysh -- I went to utilize this component in another project and it seemed overly complex considering it requires a connection to the created RDS cluster. I ended up creating a much smaller and simpler component. That is fine, but I'm wondering if we want to open source this as is? I'm thinking it would be smart to pull out the `postgres_*` resources to another component. Any strong opinions on this one?

PR #314, opened 2021-04-15
feat: adds the vpc-* components

Gowiem:
## what
* Adds the vpc-flow-logs-bucket component
* Adds the vpc-peering component

## why

### DIFF
--- a/modules/vpc-peering/README.md
+++ b/modules/vpc-peering/README.md
@@ -1,0 +1,211 @@
+# Component: `vpc-peering`
+
+This component is responsible for creating a peering connection between two VPCs existing in different AWS accounts.
+
+## Usage
+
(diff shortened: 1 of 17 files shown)


modules/vpc-peering/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+`stacks/ue1-sandbox.yaml`:
+
+```yaml
+import:
+  - vpc-peering/vpc-peering-defaults
+```
+
+
+## Legacy Account Configuration
```
Gowiem, line 33: These docs were carried over from a customer and I've anonymized them. Worth considering if they'll be useful for others or not.

nitrocode (approved):

PR #315, opened 2021-04-15
feat: adds config-bucket + s3-bucket components

Gowiem:
## what
* Adds config-bucket
* Adds s3-bucket

## why


Gowiem:
Old as dirt!

PR #316, opened 2021-04-15
feat: upstream compliance component

Gowiem:
## what
* Upstreams the compliance component

## why


Gowiem:
Old as dirt!

PR #319, opened 2021-04-16
feat: adds elasticache-redis

Gowiem:
## what
* Adds elasticache-redis component

## why

### DIFF
--- a/modules/elasticache-redis/remote-state.tf
+++ b/modules/elasticache-redis/remote-state.tf
@@ -1,0 +1,30 @@
+module "vpc" {
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "0.13.0"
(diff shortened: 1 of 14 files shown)


modules/elasticache-redis/remote-state.tf
```diff
@@ -0,0 +1,30 @@
...
+
+module "eks" {
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "0.13.0"
+
+  stack_config_local_path = "../../../stacks"
+  component               = "eks"
+
+  context = module.this.context
+}
```
osterman, line 19: This is going to bite us... since we also support ECS. Anyways, guess we'll need to generalize this later.

osterman (approved):

PR #321, opened 2021-04-22
chore: updates docs to use new vars in spacelift-worker-image

Gowiem:
## what
* Updates documentation to use new variables added to spacelift-worker-image in https://github.com/spacelift-io/spacelift-worker-image/pull/3
* Removes patch file workflow

## why
* Should be easier to maintain / build AMIs for spacelift now

## references
* This is dependent on https://github.com/spacelift-io/spacelift-worker-image/pull/3 merging first


modules/spacelift/README.md
```diff
@@ -97,20 +97,23 @@ chmod +x rootfs/usr/local/bin/spacelift*
...
-
-3. Run `packer build spacelift.pkr.hcl` against the account Spacelift will be deployed to
-4. Take note of the AMI created and use it to configure the `spacelift-worker-pool` component
+1. Grab the latest AL2 AMI ID:
+```bash
+aws ec2 describe-images --region $YOUR_TARGET_REGION --owners amazon --filters "Name=name,Values=amzn2-ami-hvm-2.0.202*-x86_64-gp2" --query 'sort_by(Images, &CreationDate)[-1]' | jq -r .ImageId
+```
+1. In the cloned repo, add a `default.hcl` file with the following content:
+```hcl
+base_ami = "$IMAGE_ID_FROM_ABOVE
```
Nuru, line 106: Please add explanation about these variables. It is not clear whether you mean to say that these variables should be set somewhere (in the environment, in a config file, etc.) or are placeholders you should replace with literal values. I am in favor of using `<placeholder>` rather than `$PLACEHOLDER` if the intent is to make a literal replacement. You can also use a real value and note both in the narrative text above and in the code via a comment that it should be replaced, e.g.

Replace `us-east-1` with the AWS region where you intend to deploy the worker pool:
```hcl 
region = "us-east-1" # replace with your target region
```

In any case, you left out a trailing quotation mark:
```suggestion
base_ami = "$IMAGE_ID_FROM_ABOVE"
```

Nuru (changes requested):

modules/spacelift/README.md
Gowiem: @Nuru good feedback -- updated 👍

Gowiem:
@Nuru friendly ping.

Gowiem:
@Nuru friendly ping.

modules/spacelift/README.md
```diff
@@ -97,20 +97,23 @@ chmod +x rootfs/usr/local/bin/spacelift*
...
-
-   Change the region as needed, but make sure you update the `base_ami` with the latest Amazon Linux 2 AMI id:
-
-   ```bash
-   aws ec2 describe-images --owners amazon --filters "Name=name,Values=amzn2-ami-hvm-2.0.202*-x86_64-gp2" --query 'sort_by(Images, &CreationDate)[-1]'
-   ```
-
-3. Run `packer build spacelift.pkr.hcl` against the account Spacelift will be deployed to
-4. Take note of the AMI created and use it to configure the `spacelift-worker-pool` component
+1. Grab the latest AL2 AMI ID:
```
nitrocode, line 100: The filter already takes care of this so no more need to pass in the hard coded ami.

nitrocode: See this page https://github.com/spacelift-io/spacelift-worker-image/blob/main/aws/README.md

Packer isn't even needed anymore using the data source in the readme since spacelift packs images for you in desired regions. They are also open to packing their images in mew regions.

Gowiem:
@nitrocode considering `packer` is no longer needed for this biz... I'm closing this PR out as this is likely legacy at this point. Please ping me if you feel differently!

PR #322, opened 2021-04-23
feat: adds a new abstract providers.tf proposal

Gowiem:
## what

* Proposes a new pattern for abstractly defining the aws provider via a common `providers.tf`

## why

* This enables usage of the following types of AWS auth for components: 
	* Environment credentials (`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`) via `iam_roles_profile_enabled = false` and not touching the other vars
	* Profile from iam-roles
	* Role ARN from iam-roles
	* Manually passed Profile (Solves the import need)
	* Manually passed Role ARN (Solves the import need)

## Notes

* This does require the iam-roles which is sitting on PR now in the `upstream-account-comps` branch: #304 
	* Tried making it work with `0.146.0`, but that is older and requires `region` which is annoying. 
	* Locally I updated to use the github URL: `git::https://github.com/cloudposse/terraform-aws-components.git//modules/account-map/modules/iam-roles/?ref=upstream-account-comps`

### DIFF
--- a/modules/vpc/README.md
+++ b/modules/vpc/README.md
@@ -45,7 +45,7 @@ components:
 
 | Name | Source | Version |
 |------|--------|---------|
-| <a name="module_iam_roles"></a> [iam\_roles](#module\_iam\_roles) | ../account-map/modules/iam-roles |  |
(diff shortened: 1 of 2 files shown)


Gowiem:
@osterman @aknysh @Nuru I'm going to close this one out as I believe we discussed that we don't want to utilize a complex pattern here and instead we'll expect folks to vendor in and handle their provider authentication mechanisms on their own.

PR #329, opened 2021-06-01
Fixes and changes to spin up a brand new atmos stack

iliazlobin:
## what
- remote_state invocation through the module, removing old code where remote state is pulled through data sources
- updated modules' versions in several places
- changes to providers.tf for bootstraping components with a direct assume role variable (we haven't figured out how to make it work through iam_roles module while we don't have any roles provisioned)
- missing and required variables that have references in module's code
- we'd be rather using roles instead of profiles when setting up AWS providers
- updated policies for ALB controller according to updated AWS docs
- created custom EKS IAM roles


modules/sso/providers.tf
```diff
@@ -1,25 +1,12 @@
...
 
-  dynamic "assume_role" {
-    for_each = module.iam_roles.org_role_arn != null ? [true] : []
-    content {
-      # `terraform import` will not use data from a data source,
-      # so on import we have to explicitly specify the role
-      role_arn = coalesce(var.import_role_arn, module.iam_roles.org_role_arn)
-    }
+  assume_role {
+    role_arn = var.aws_assume_role_arn
```
Nuru, line 5: All the changes to this file should be reverted

modules/vpc-flow-logs-bucket/providers.tf
```diff
@@ -2,17 +2,11 @@ provider "aws" {
   region = var.region
 
   assume_role {
-    role_arn = coalesce(var.import_role_arn, module.iam_roles.terraform_role_arn)
+    role_arn = var.aws_assume_role_arn
```
Nuru, line 5: All the changes to this file should be reverted

modules/vpc/providers.tf
```diff
@@ -2,12 +2,14 @@ provider "aws" {
...
   assume_role {
+    # `terraform import` will not use data from a data source,
+    # so on import we have to explicitly specify the role
     role_arn = coalesce(var.import_role_arn, module.iam_roles.terraform_role_arn)
   }
 }
 
 module "iam_roles" {
-  source  = "../account-map/modules/iam-roles"
+  source = "../account-map/modules/iam-roles"
```
Nuru, line 12: Revert

Nuru (changes requested):

Nuru:
I rejected the changes to the providers for the bootstrap components because they create or modify IAM roles and/or authentication so you cannot use the roles they create to create them (which is why they are called "bootstrap" components).

modules/eks/versions.tf
```diff
@@ -12,7 +12,7 @@ terraform {
     }
     kubernetes = {
       source  = "hashicorp/kubernetes"
-      version = "~> 1.13"
+      version = "~> 2.2.0"
```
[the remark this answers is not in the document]
iliazlobin: We were able to deploy the eks component with version 2.2.0 quite successfuly. Could it be that the problem was related to some of the previous versions of the k8s provider (2.1.x) that's been fixed in 2.2.0?

sgtoj:
Maybe I am mistaken, but I don think you meant to commit...
```
modules/tfstate-backend/gbl-root-tfstate-backend.planfile
```

nitrocode:
I think this was included on accident as well `gbl-root-tfstate-backend.tfplan`.

modules/efs/main.tf
```diff
@@ -6,7 +6,7 @@ locals {
 }
 
 module "efs" {
-  source = "git::https://github.com/cloudposse/terraform-aws-efs.git?ref=tags/0.21.0"
+  source = "git::https://github.com/cloudposse/terraform-aws-efs.git?ref=tags/0.30.1"
```
nitrocode, line 9: nit: We should use the more popular `source`

```suggestion
  source  = "cloudposse/efs/aws"
  version = "0.30.1"
```

modules/efs/remote-state.tf
```diff
@@ -9,6 +9,7 @@ module "vpc" {
 }
 
 module "eks" {
+  count = var.use_eks_security_group ? 1 : 0
```
nitrocode, line 12:
```suggestion
  count   = var.use_eks_security_group ? 1 : 0
```

Benbentwo:
Closed this as it has become stale, most of the components referenced by this branch and PR have been superseded or updated.

PR #335, opened 2021-07-01
fix: marks dns-delegated acm_ssm_parameter output as sensitive

Gowiem:
## what
* Marks the `acm_ssm_parameter` output in `dns-delegated/acm.tf` as `sensitive = true`

## why
* Fixes issue with 0.15+ not liking that output being in the clear: 
CleanShot 2021-07-01 at 17 00 22

### DIFF
--- a/modules/dns-delegated/acm.tf
+++ b/modules/dns-delegated/acm.tf
@@ -49,4 +49,5 @@ output "certificate" {
 output "acm_ssm_parameter" {
   value       = aws_ssm_parameter.acm_arn
   description = "The SSM parameter for the ACM cert."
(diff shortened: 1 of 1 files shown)


osterman:
Why is this sensitive?

Gowiem:
@osterman I think the aws provider marks the full SSM PStore object as sensitive if it's output'd in full. See the screenshot in description. This was the path forward to get around that issue, but if you want to try something else instead then let me know.

Gowiem:
@osterman friendly ping.

nitrocode (approved):

PR #336, opened 2021-07-06
Upstream `zscaler` component

nitrocode:
## what
* Upstream `zscaler` component

## why
* Others can now provision ZScaler Private Access Connector instances on Amazon Linux 2

## references
N/A

## notes

To pass pre-commit hook

```shell
tfenv install 0.13.5
make readme
pre-commit run --show-diff-on-failure --color=always --all-files
git add modules/
```

### DIFF
--- a/modules/account-map/README.md
+++ b/modules/account-map/README.md
@@ -38,7 +38,6 @@ components:
 | Name | Version |
 |------|---------|
 | <a name="provider_aws"></a> [aws](#provider\_aws) | ~> 3.32 |
-| <a name="provider_terraform"></a> [terraform](#provider\_terraform) | n/a |
 
 ## Modules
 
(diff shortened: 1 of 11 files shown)


modules/zscaler/remote-state.tf
```diff
@@ -0,0 +1,19 @@
...
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "0.17.0"
+
+  stack_config_local_path = "../../../stacks"
+  component               = "vpc"
+
+  context = module.this.context
+}
+
+module "eks" {
```
jamengual, line 11: this is specific to customers, maybe it should not be here?

nitrocode: I think this is still required for using the sg of the eks in the zscaler ec2 instance

jamengual (approved):

PR #341, opened 2021-07-09
Fix: Bastion Component (module delegation, private subnet deployment)

korenyoni:
## what
* Delegate EIP, EBS volume, Route53 record to `ec2-bastion` module rather than component
* Support deployment to private subnet (bump module version)
* Fix IAM Role not being attached (bump module version)
* Add documentation for additional use cases
* Change name of route53 variable to make it less opinionated (i.e. not necessarily required for a vanity domain); fix variable description punctuation.
* Remove `user_data` variable as it conflicts with `user_data_base64` which we pass in.
* Change variable name for ssh key enablement variable, ensure that component documentation reflects the fact that the defaults are suitable for an SSM bastion and not an SSH bastion; ensure `key-pair `module is disabled when ssh key is disabled.

## why
* Deployment to private subnets was not working previously
* The component was managing its own EIP, EBS Volume, Route53 record resources rather than delegating that work to the module.

## references
* https://github.com/cloudposse/terraform-aws-ec2-bastion-server/pull/80
* https://github.com/cloudposse/terraform-aws-ec2-bastion-server/pull/82
* https://github.com/cloudposse/terraform-aws-ec2-bastion-server/pull/83


modules/bastion/main.tf
```diff
@@ -34,20 +32,11 @@ locals {
   })
 }
 
-data "aws_route53_zone" "vanity" {
-  count = local.route53_enabled ? 1 : 0
+data "aws_route53_zone" "route53_zone" {
+  count = try(length(var.vanity_domain), 0) > 0 ? 1 : 0
   name  = var.vanity_domain
```
nitrocode, line 37: The vanity domain should be renamed to something like `var.vanity_domain_zone` or something similar.

modules/bastion/main.tf
```diff
@@ -34,20 +32,11 @@ locals {
...
-resource "aws_route53_record" "default" {
-  count   = local.route53_enabled ? 1 : 0
-  zone_id = join("", data.aws_route53_zone.vanity.*.zone_id)
-  name    = "${var.custom_bastion_hostname}.${var.vanity_domain}"
-  type    = "A"
-  ttl     = "300"
-  records = aws_eip.static.*.public_ip
-}
-
 module "aws_key_pair" {
```
nitrocode, line 40: With ssm, the key pair is not needed. It would be nice to disable this as a default and only enabled if explicitly needed. Can be done in a separate pr.

modules/bastion/main.tf
```diff
@@ -100,41 +84,22 @@ module "ec2_bastion" {
...
+  host_name          = var.custom_bastion_hostname
 
-  ebs_block_device_volume_size = 0
+  ebs_block_device_volume_size = var.ebs_block_device_volume_size
   ebs_delete_on_termination    = var.ebs_delete_on_termination
   user_data_base64             = join("", data.cloudinit_config.config.*.rendered)
   security_group_rules         = var.security_group_rules
   key_name                     = module.aws_key_pair.key_name
   instance_profile             = join("", aws_iam_instance_profile.default.*.name)
+  user_data                    = var.user_data
```
nitrocode, line 97: If this is passed in, it will disable `user_data_base64`. Might be worth documenting this. I believe the upstream module variables do document this. Could be done in a follow up pr.

nitrocode (approved):

modules/bastion/main.tf
osterman: Vanity domain is how the customer chose to use it. But I think it's too opinionated. Let's just call it dns zone.

osterman: See our route 53 module. Use that instead.

korenyoni: @osterman 

> See our route 53 module. Use that instead.

Do you mean the variable description in that module or the module itself? I'm assuming the former, because the `ec2-bastion-server` module manages Route53 resources itself.

modules/bastion/main.tf
korenyoni: There is a definite use case for using the key pair, as SSM is toggleable and hence in some cases a traditional `sshd` bastion will be used in lieu of an SSM bastion, but I will disable this module by default

modules/bastion/main.tf
korenyoni: Will account for this

modules/bastion/main.tf
korenyoni: resolved

modules/bastion/main.tf
korenyoni: Resolved

modules/bastion/main.tf
korenyoni: resolved

osterman (approved):

PR #346, opened 2021-08-08
Update efs to use remote-state.tf

sboardwell:
## what

- Update efs component to use `remote-state.tf`

## why

- To bring it inline with the new atmos setup.


modules/efs/main.tf
```diff
@@ -1,8 +1,8 @@
 locals {
-  eks_cluster_managed_security_group_id = data.terraform_remote_state.eks.outputs.eks_cluster_managed_security_group_id
-  private_subnet_ids                    = data.terraform_remote_state.vpc.outputs.private_subnet_ids
-  vpc_id                                = data.terraform_remote_state.vpc.outputs.vpc_id
-  zone_id                               = data.terraform_remote_state.dns_delegated.outputs.default_dns_zone_id
+  eks_cluster_managed_security_group_id = module.eks.outputs.eks_cluster_managed_security_group_id
```
nitrocode, line 2: I'm newer versions, we have a toggle for eks security groups because some teams don't have eks deployed in every environment that requires efs

sboardwell: Thanks @nitrocode, how does the toggle look (what is the toggle based upon)?

Do you perhaps have an example for me?

nitrocode: It hasn't been upstreamed yet. It's something like this but it should be in the appropriate files.

```hcl
variable "use_eks_security_group" {
  type        = bool
  description = "Use the eks default security group"
  default     = false
}

locals {
  existing_security_groups = var.use_eks_security_group ? [module.eks[0].outputs.eks_cluster_managed_security_group_id] : []
}

module "eks" {
  count = var.use_eks_security_group ? 1 : 0
  # ...
}

module "efs" {
  # ...
  security_groups                 = local.existing_security_groups
  # ...
}
```

sboardwell: Done

sboardwell: btw what has changed in the `eks` module to make it an array? `module.eks[0]` gives us an error atm.

sboardwell: The remote-state doesn't suggest anything...

```hcl
module "eks" {
  version = "0.17.0"

  component               = "eks"
  stack_config_local_path = "../../../stacks"

  context = module.this.context
}
```

modules/efs/main.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  eks_cluster_managed_security_group_id = module.eks.outputs.eks_cluster_managed_security_group_id
+  existing_security_groups = var.use_eks_security_group ? [module.eks[0].outputs.eks_cluster_managed_security_group_id] : []
```
nitrocode, line 2: Is the count logic set on the eks remote state? Otherwise the 0 index would throw an error.

sboardwell: Ah, yes. That's what I was seeing with https://github.com/cloudposse/terraform-aws-components/pull/346#discussion_r684980771

Thanks. Works now.

sboardwell:
@nitrocode - is this okay now?

PR #348, opened 2021-08-12
feat(component): Add Client VPN Terraform Module 

r351574nc3:
## what
* New component/module for creating client-vpn setup on AWS

## why
* Use terraform to manage AWS EC2 VPN through Infrastructure as code
* Simplify/streamline use of `aws ec2 export-client-vpn-client-configuration` when provisioning VPN for users


r351574nc3:
There's a new component

PR #355, opened 2021-10-05
Feat: add `alb-controller` EKS component

korenyoni:
## what
* Add `alb-controller` EKS component

## why
* [alb-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller) manages ALBs and NLBs based on Ingress and Service k8s annotations, and makes the transparent management of public-facing services via k8s objects trivial.

## references
* N/A


dylanbannon:
Two questions:
1. Can you get the `pre-commit` check to pass? This might involve installing pre-commit hooks locally.
2. Just to confirm, have you deployed this component for a customer already?

PR #358, opened 2021-10-05
Feat: add `aws-node-termination-handler` EKS component

korenyoni:
## what
* Add `aws-node-termination-handler` EKS component.

## why
* By default, `aws-node-termination-handler` monitors the EC2 IMDS endpoint for scheduled maintenance events, spot instance termination events, and rebalance recommendation events, and drains and/or cordons nodes upon such events. Using this component results in more robust EKS architecture.

## references
* https://github.com/aws/aws-node-termination-handler


dylanbannon:
@korenyoni Do we need to make this component compatible with pre-1.0 versions of Terraform? If so, can you indicate how far back it needs to remain compatible by adding the `terraform/[version` tage to this PR? (Assigning a `terraform/[version]` tag should lead to pre-commit checks rerunning, per-commit autocommitting any changes, and the pre-commit checks passing.)

PR #373, opened 2021-11-03
Upgrade EKS (blocked on spot)

nitrocode:
## what
* Upgrade EKS
* Disable Spotinst support

## why
* Latest eks, node group, context
* Tenant support
* Any support of Spotinst currently requires valid Spotinst authentication credentials, even if no Spotinst resources are created or managed

## references
* https://github.com/cloudposse/terraform-aws-eks-cluster 0.43.2
* https://github.com/cloudposse/terraform-aws-eks-node-group 0.26.2
* https://github.com/spotinst/terraform-provider-spotinst/issues/253


Nuru (changes requested):
We have to have a better solution than this for Spotinst support or not support. Commenting out half the support code is not a viable option for a public component.

Nuru (changes requested):
@osterman I think what we should do is move the Spotinst support to a sub-module (including moving the Spotinst provider into the module), like we did with `node_group_by_region`, and provide an example invocation in a file like `spotinst.tf.example`. This way, users can just copy the example file into a live file and get the updates as they are released, since the updates will likely be to the sub-module and not the invocation.

PR #378, opened 2021-11-24
feat: Upstream `datadog-agent`

nitrocode:
## what
* Upstream `datadog-agent`

## why
* Upstream

## references
N/A


korenyoni (approved):
LGTM

nitrocode:
This is old but we can always put in a follow up PR

PR #384, opened 2022-01-05
Feat: `s3-bucket` Upstream

milldr:
## what
* upstreamed `s3-bucket`

## why
* updating component and context for client changes
* provided examples for using virtual component inheritance


modules/s3-bucket/README.md
```diff
@@ -6,53 +6,111 @@ This component is responsible for provisioning S3 buckets.
 
 **Stack Level**: Regional
 
-Here's an example snippet for how to use this component.
+Atmos now supports virtual component inheritance. We can leverage this feature to keep bucket configuration completely `D.R.Y.`.  Here's an example snippet with multiple `s3-bucket` types:
```
korenyoni, line 9: I would say this is a nice explanation but the comment is very specific about atmos at this point in time. Maybe just say this component leverages virtual component inheritance? Not a dealbreaker, just a suggestion.

korenyoni (approved):
Approved with minor comment

modules/s3-bucket/README.md
milldr: good point. we should note that this component will not work unless atmos is updated to at least version `v1.3.12`

korenyoni (approved):

PR #387, opened 2022-01-26
Echo-server ALB compatibility

dylanbannon:
## what
* Adding the ability to choose between NGiNX and ALB backends for the `echo-server` component.

## why
* Sometimes, clusters use ALB instead of NGiNX.


dylanbannon:
@Nuru can you confirm this works with ALB?

max-lobur (approved):

modules/echo-server/charts/echo-server/templates/ingress.yaml
```diff
@@ -1,55 +1,55 @@
...
+    kubernetes.io/ingress.class: {{ .Values.ingress.nginx.class }}
+      {{- if (index .Values.ingress.nginx "tls_certificate_cluster_issuer") }}
+    cert-manager.io/cluster-issuer: {{ .Values.ingress.nginx.tls_certificate_cluster_issuer }}
+      {{- end }}
+    {{- else if eq (printf "%v" .Values.ingress.alb.enabled) "true" }}
+    kubernetes.io/ingress.class: {{ .Values.ingress.alb.class }}
+    alb.ingress.kubernetes.io/load-balancer-name: {{ index .Values.ingress.alb "load_balancer_name" | default "k8s-common" }}
+    alb.ingress.kubernetes.io/group.name: {{ index .Values.ingress.alb "group_name" | default "common" }}
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=false
```
max-lobur, line 25: This kind of annotations need to be moved to a "default" alb created within the alb controller component. After that removed from here. They are not allowed to be changed per ingress, they are global to an ALB (global among users of a single alb.ingress.kubernetes.io/load-balancer-name).

aknysh: besides this comment, everything else looks good

aknysh (commented):
please see comments

modules/echo-server/charts/echo-server/templates/ingress.yaml
Nuru: @max-lobur @aknysh We do not have a standard mechanism for parallel configuration of Ingresses in that way, and are not about to develop one today for the open source component. Can the `load-balancer-attributes` annotation be safely left off? If not, I say we stick with this and require `echo-server` to use its own Load Balancer if needed.

Nuru:
Ignore the pre-commit check failure: it is complaining that the template in the Helm chart are not valid Kubernetes resources (which of course they are not, they are Go templates).

Nuru (changes requested):
Please verify this still works with Nginx after my changes

Nuru (approved):

modules/echo-server/charts/echo-server/templates/ingress.yaml
max-lobur: Agreed. Securing annotations should be a separate PR

PR #391, opened 2022-01-31
Feat: Update `account` and `account-map` components

### LINKED ISSUE #390: `account` component incomaptible with `account-map`
## Describe the Bug

The most recent update of `account-map` (#363) is using an output of `account` called `account_info_map` that is not present in the current published version of the component.

## Expected Behavior

```
atmos terraform plan -s gbl-root account-map
```

works

## Steps to Reproduce
1. Use the latest published `account` component and apply it
2. Use the latest published `account-map` component plan/apply it
3. See error 
```
│ Error: Unsupported attribute
│ 
│   on main.tf line 12, in locals:
│   12:   account_info_map = module.accounts.outputs.account_info_map
│     ├────────────────
│     │ module.accounts.outputs is object with 21 attributes
│ 
│ This object does not have an attribute named "account_info_map".
```

## Additional Context

I know that the latest components are not always published, but it would be great to have them at least work together. I'm happy to help get them merged into the main branch of this repository.

korenyoni:
## what
* Update `account` and `account-map` components to their latest versions.

## why
* These two components are out of date — also, the current version `account-map` component is not compatible with `account`.
* Update `context.tf` in `account`.

## references
* Closes #390


modules/account-map/README.md
```diff
@@ -14,9 +14,6 @@ components:
     account-map:
       vars:
         enabled: true
-        # If this is true, profile must be provided instead of role_arn in each backend.tf.json
-        # If this is false, role_arn must be provided instead of profile in each backend.tf.json
-        profiles_enabled: false
```
nitrocode, line 19: Why is this removed ?

modules/account-map/modules/roles-to-principals/variables.tf
```diff
@@ -17,8 +17,8 @@ variable "privileged" {
 
 variable "global_tenant_name" {
   type        = string
-  description = "Global tenant name"
-  default     = null
+  description = "The tenant name used for organization-wide resources"
+  default     = "mgmt"
```
nitrocode, line 21: Why does this default to mgmt ?

modules/account/README.md
```diff
@@ -45,13 +47,13 @@ components:
...
-                - name: platformprod
+                - name: prodplatform
                   tags:
                     eks: true
-                - name: platformdev
+                - name: devplatform
                   tags:
                     eks: true
-                - name: platformstaging
+                - name: stageplatform
```
nitrocode, line 56: Why are these modified ?

modules/account-map/README.md
korenyoni: Restored it immediately and forgot to push

modules/account-map/modules/roles-to-principals/variables.tf
korenyoni: Otherwise we'd always have to supply this value when instantiating the submodule. But we can make it `null` if you want.

modules/account/README.md
korenyoni: Copypasta — I can restore

nitrocode:
@korenyoni check yaml is failing the pre-commit check

modules/account/README.md
korenyoni: Tbh this stays more consistent with `proddata` — I think it's better actually

korenyoni:
> @korenyoni check yaml is failing the pre-commit check

Seems unrelated to this changeset but will take a look.

korenyoni:
> @korenyoni I think we should keep it a habit to keep it green. Unsure why it's not a requirement to merge the PR.

Yeah for sure — I want to keep it green before merging. But I'm scratching my head as to why it started failing from this changeset.

korenyoni:
Yeah... not sure why the check failed. We are specifically excluding these manifests:

https://github.com/cloudposse/terraform-aws-components/blob/046819528e0b58db00c53b1d36b9bf00c53125d4/.pre-commit-config.yaml#L10

```
while parsing a flow node
did not find expected node content
  in "modules/echo-server/charts/echo-server/templates/deployment.yaml", line 6, column 7
while parsing a flow node
did not find expected node content
  in "modules/echo-server/charts/echo-server/templates/ingress.yaml", line 1, column 3
while parsing a flow node
did not find expected node content
  in "modules/echo-server/charts/echo-server/templates/service.yaml", line 6, column 7

```

nitrocode:
Not sure but it did work in the previous PR https://github.com/cloudposse/terraform-aws-components/pull/387

nitrocode:
@korenyoni it looks like this PR wasn't branched off of the latest default branch.

https://github.com/cloudposse/terraform-aws-components/blob/c64ffe9b11020682c34cb9741cf20a023fc24794/.pre-commit-config.yaml#L6-L10

Merge in the master branch and push and it should resolve the issue.

korenyoni:
> @korenyoni it looks like this PR wasn't branched off of the latest default branch.
> 
> https://github.com/cloudposse/terraform-aws-components/blob/c64ffe9b11020682c34cb9741cf20a023fc24794/.pre-commit-config.yaml#L6-L10
> 
> Merge in the master branch and push and it should resolve the issue.

Ah makes sense. Oversight on my part as I did this PR quite hastily.

Done

modules/account/outputs.tf
```diff
@@ -18,14 +18,26 @@ output "organizational_unit_ids" {
   description = "List of Organizational Unit IDs"
 }
 
+output "account_info_map" {
+  value       = local.account_info_map
+  description = <<-EOT
+    Map of account names to
+      eks: boolean, account hosts at least one EKS cluster
+      id: account id (number)
+      stage: (optional) the account "stage"
```
Nuru, line 27: `stage` is not optional

modules/account-map/modules/iam-roles/main.tf
```diff
@@ -14,10 +14,10 @@ module "account_map" {
 
   component               = "account-map"
   privileged              = var.privileged
-  tenant                  = var.global_tenant_name
   environment             = var.global_environment_name
-  stack_config_local_path = "../../../stacks"
+  stack_config_local_path = var.stack_config_local_path
   stage                   = var.root_account_stage_name
+  tenant                  = var.root_account_tenant_name
```
Nuru, line 20: I would prefer to leave this as 
```suggestion
  tenant                  = var.global_tenant_name
```

modules/account-map/modules/iam-roles/variables.tf
```diff
@@ -4,12 +4,6 @@ variable "privileged" {
   default     = false
 }
 
-variable "global_tenant_name" {
-  type        = string
-  description = "Global tenant name"
-  default     = null
-}
-
```
Nuru, line 12: I would prefer to leave this. So far we haven't had a client for whom this was not sufficient.

modules/account-map/modules/iam-roles/variables.tf
```diff
@@ -21,3 +15,24 @@ variable "root_account_stage_name" {
...
+
+variable "root_account_tenant_name" {
+  type        = string
+  description = <<-EOT
+  The tenant name for the root account.
+
+  If the `tenant` label is not used, set this to `null`.
+  EOT
+  default     = "mgmt"
+}
```
Nuru, line 27: I would prefer to have and use just the one `global_tenant_name`. So far that has been sufficient.

modules/account-map/modules/roles-to-principals/variables.tf
Nuru: Because this is a sub-module, we cannot supply a `default.auto.tfvars`, and we do not want to have to chain this through every client module, so in general this value should be directly edited/changed by the user here and in `../iam-roles/variables.tf`.

PR #395, opened 2022-03-02
update github-runners to support using registration token directly

mcalhoun:
## what

Github now supports obtaining a scoped (Organization or Repository) registration token directly through the UI, which would eliminate the need for any kind of long-lived credentials to be stored. It would be great if we could specify the registration token rather than a PAT and the actions-runner-controller could just use the registration token directly rather than using the API to obtain one.

## why

Currently, the component supports providing a Github Personal Access Token (PAT) and uses that token to call the Github API to request a Runner registration token. Using this method requires users to create and store a privileged (Admin scoped) token that has more permissions than it requires.

## references
* Additional information in the [RefArch documentation](https://cloudposse.atlassian.net/wiki/spaces/REFARCH/pages/1180303385/github-runners)

### DIFF
--- a/modules/github-runners/README.md
+++ b/modules/github-runners/README.md
@@ -2,6 +2,10 @@
 
 This component is responsible for provisioning EC2 instances for GitHub runners.
 
+## Pre-Requisites
+
+Before using this component, you must obtain a Github Runner registration token and store it in SSM parameter store. Please see the [Reference Architecture documentation](https://cloudposse.atlassian.net/wiki/spaces/REFARCH/pages/1180303385/github-runners) for details on how to do this.
+
 ## Usage
 
 **Stack Level**: Regional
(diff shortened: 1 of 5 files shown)


modules/github-runners/README.md
```diff
@@ -2,6 +2,10 @@
 
 This component is responsible for provisioning EC2 instances for GitHub runners.
 
+## Pre-Requisites
+
+Before using this component, you must obtain a Github Runner registration token and store it in SSM parameter store. Please see the [Reference Architecture documentation](https://cloudposse.atlassian.net/wiki/spaces/REFARCH/pages/1180303385/github-runners) for details on how to do this.
+
```
htplbc, line 8: Do we normally reference atlassian wiki pages on this public repo?

htplbc (approved):
lg!

PR #396, opened 2022-03-04
feat: upstream `opsgenie-team`

Benbentwo:
## what
* A Component that manages an opsgenie team and its corresponding resources

## why
* The OpsGenie Team is a core component that needs many backing resources spun up. This creates those sane defaults

### DIFF
--- a/modules/opsgenie-team/modules/integration/main.tf
+++ b/modules/opsgenie-team/modules/integration/main.tf
@@ -103,9 +103,59 @@ resource "opsgenie_integration_action" "datadog" {
 
       conditions {
         # not `source` as described in the TF docs
-        field          = "source_type_name"
+        field          = "event_type"
         operation      = "equals"
-        expected_value = "Monitor Alert"
+        expected_value = "query_alert_monitor"
+      }
+
+    }
+
+    responders {
+      id   = local.team_id
+      type = "team"
+    }
+  }
+  
+    create {
+    ignore_responders_from_payload = true
+
+    name = "Create Synthetic Alert"
+    # Only add the Statuspage component and incident ID tags.
+    # If all of dd_tags is added, then the alert will be polluted with probe_ tags, which are not necessary in OpsGenie,
+    # and also cause the alert's tag count to exceed the limit (20). This then prevents the Statuspage integration from
+    # appending the incident_id tag, which is required for bidirectional OpsGenie <-> Statuspage communication.
+    tags = [
+      "{{ dd_tags.substringBetween(\"cmp_\",\",\") }}",
+      "{{ dd_tags.substringBetween(\"incident_id:\",\",\") }}"
+    ]
+    user          = "Datadog"
+    note          = "{{note}}"
+    alias         = "{{alias}}"
+    source        = "{{source}}"
+    message       = "{{monitor_name.substringAfter(\"[Synthetics]\")}}"
+    entity        = "{{entity}}"
+    alert_actions = []
+
+    description = "{{templated_message.substringBefore(\"@\")}}\n"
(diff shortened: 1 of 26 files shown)


modules/opsgenie-team/modules/integration/main.tf
```diff
@@ -0,0 +1,194 @@
...
+
+      # We purposely do not check for tenant/stage since this is controlled by the Datadog organizations
+      # and how the datadog organizations are configured. 1 per tenant, 1 per account, single org, etc.
+
+      conditions {
+        # not `source` as described in the TF docs
+        field          = "source_type_name"
+        operation      = "equals"
+        expected_value = "Monitor Alert"
+      }
```
korenyoni, line 109: Do not create Datadog Synthetics when only monitors are to be created.

```suggestion
      conditions {
        # not `source` as described in the TF docs
        field          = "event_type"
        operation      = "equals"
        expected_value = "query_alert_monitor"
      }
```

modules/opsgenie-team/modules/integration/main.tf
```diff
@@ -0,0 +1,194 @@
...
+
+    }
+
+    responders {
+      id   = local.team_id
+      type = "team"
+    }
+  }
```
korenyoni, line 117: Allow creating Datadog Synthetics Alerts

```suggestion
  }
  
    create {
    ignore_responders_from_payload = true

    name = "Create Synthetic Alert"
    # Only add the Statuspage component and incident ID tags.
    # If all of dd_tags is added, then the alert will be polluted with probe_ tags, which are not necessary in OpsGenie,
    # and also cause the alert's tag count to exceed the limit (20). This then prevents the Statuspage integration from
    # appending the incident_id tag, which is required for bidirectional OpsGenie <-> Statuspage communication.
    tags = [
      "{{ dd_tags.substringBetween(\"cmp_\",\",\") }}",
      "{{ dd_tags.substringBetween(\"incident_id:\",\",\") }}"
    ]
    user          = "Datadog"
    note          = "{{note}}"
    alias         = "{{alias}}"
    source        = "{{source}}"
    message       = "{{monitor_name.substringAfter(\"[Synthetics]\")}}"
    entity        = "{{entity}}"
    alert_actions = []

    description = "{{templated_message.substringBefore(\"@\")}}\n"
    extra_properties = {
      "Event Url" : "{{event_url}}"
    }

    filter {
      type = "match-all-conditions"

      conditions {
        # not `actions` as described in the TF docs
        field          = "action"
        operation      = "equals"
        expected_value = "create"
      }

      conditions {
        # not `source` as described in the TF docs
        field          = "event_type"
        operation      = "equals"
        expected_value = "synthetics_alert"
      }

    }

    responders {
      id   = local.team_id
      type = "team"
    }
  }
```

modules/opsgenie-team/versions.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_version = ">= 1.0.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.0"
```
korenyoni, line 7: Are we ready to pin any components past 3.x?

```suggestion
      version = "~> 3.0"
```

korenyoni (commented):
See comments

modules/opsgenie-team/versions.tf
Benbentwo: Yeah only our s3 component really prevented us

Benbentwo: meaning this can be pinned to 4.0

korenyoni (approved):

PR #404, opened 2022-04-16
update (account-map/modules/iam-roles): update remote state to version 0.22.1

mikedizon:
## what
* account-map/modules/iam-roles, account-map/remote-state - updated remote-state version to 0.22.1

## why
* updating remote state to 0.22.1 from 0.19.0 resolves issue with latest version of atmos

### DIFF
--- a/.github/auto-release.yml
+++ b/.github/auto-release.yml
@@ -4,30 +4,36 @@ version-template: '$MAJOR.$MINOR.$PATCH'
 version-resolver:
   major:
     labels:
-      - 'major'
+    - 'major'
   minor:
     labels:
-      - 'minor'
-      - 'enhancement'
+    - 'minor'
+    - 'enhancement'
   patch:
     labels:
-      - 'patch'
-      - 'fix'
-      - 'bugfix'
-      - 'bug'
-      - 'hotfix'
+    - 'auto-update'
+    - 'patch'
+    - 'fix'
+    - 'bugfix'
+    - 'bug'
+    - 'hotfix'
+    - 'no-release'
   default: 'minor'
 
 categories:
-  - title: '🚀 Enhancements'
-    labels:
-      - 'enhancement'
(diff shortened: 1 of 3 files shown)


modules/account-map/modules/iam-roles/outputs.tf
```diff
@@ -8,7 +8,7 @@ output "terraform_profile_name" {
 
 output "org_role_arn" {
   value = local.account_name == module.account_map.outputs.root_account_account_name ? null : format(
-    "arn:aws:iam::%s:role/OrganizationAccountAccessRole",
+    "${var.arn_format}:iam::%s:role/OrganizationAccountAccessRole",
```
aknysh, line 11: @mikedizon  thanks for the PR.

Instead of the new variable, please use:

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition

similar to:
https://github.com/cloudposse/terraform-aws-ecr/blob/58ca80423fc03e97846fbba44efb6aca9b4a3d6b/main.tf#L90

https://truss.works/blog/adventures-in-govcloud

aknysh (changes requested):
@mikedizon thanks, please see comments

mikedizon:
> @mikedizon thanks, please see comments

@aknysh i'll have to add the aws provider to use the `aws_partition` datasource. is that acceptable?

modules/account-map/modules/iam-roles/outputs.tf
mikedizon: @aknysh tried this approach last night. there seems to be a dependency cycle issue when using the `aws_partition` data source. the aws provider in the parent module references the output being modified in the `iam_roles` submodule. 

https://github.com/cloudposse/terraform-aws-components/blob/master/modules/account-settings/providers.tf#L7

adding an aws provider in the `iam_roles` submodule itself fixes this but we'd have to pass a variable for the region.

PR #405, opened 2022-04-26
update(account-map): update remote-state version

mikedizon:
## what
* account-map/modules/iam-roles, account-map/remote-state - updated remote-state version to 0.22.1

## why
* updating remote state to 0.22.1 from 0.19.0 resolves issue with latest version of atmos

### DIFF
--- a/modules/account-map/modules/iam-roles/main.tf
+++ b/modules/account-map/modules/iam-roles/main.tf
@@ -10,14 +10,13 @@ module "always" {
 
 module "account_map" {
   source  = "cloudposse/stack-config/yaml//modules/remote-state"
-  version = "0.19.0"
+  version = "0.22.1"
 
-  component               = "account-map"
-  privileged              = var.privileged
-  environment             = var.global_environment_name
-  stack_config_local_path = var.stack_config_local_path
-  stage                   = var.root_account_stage_name
-  tenant                  = var.root_account_tenant_name
+  component   = "account-map"
+  privileged  = var.privileged
+  environment = var.global_environment_name
+  stage       = var.root_account_stage_name
+  tenant      = var.root_account_tenant_name
 
   context = module.always.context
 }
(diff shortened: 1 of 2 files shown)


modules/account-map/remote-state.tf
```diff
@@ -1,10 +1,9 @@
 module "accounts" {
   source  = "cloudposse/stack-config/yaml//modules/remote-state"
-  version = "0.19.0"
+  version = "0.22.1"
 
-  component               = "account"
-  privileged              = true
-  stack_config_local_path = "../../../stacks"
```
Gowiem, line 7: I'm seeing `stack_config_local_path` as required in `0.22.1` -- https://github.com/cloudposse/terraform-yaml-stack-config/tree/0.22.1#input_stack_config_local_path

Have you tested that removing this works?

modules/account-map/modules/iam-roles/main.tf
```diff
@@ -10,14 +10,13 @@ module "always" {
 
 module "account_map" {
   source  = "cloudposse/stack-config/yaml//modules/remote-state"
-  version = "0.19.0"
+  version = "0.22.1"
 
-  component               = "account-map"
-  privileged              = var.privileged
-  environment             = var.global_environment_name
-  stack_config_local_path = var.stack_config_local_path
```
Gowiem, line 18: Same issue as below regarding this attribute.

modules/account-map/remote-state.tf
mikedizon: @Gowiem it works without it. there are other instances in the repo where 0.22.1 does not include `stack_config_local_path`

Gowiem: Ah -- Apologies @mikedizon, you're correct. I'm missing the fact that this is the `remote-state` child submodule, not the root child module. 

Let me re-review...

Gowiem (dismissed):
Looks good 👍

Gowiem:
@mikedizon I'm checking internally on the pre-commit stuff. Ping me if you don't hear from me in a day or two. 👍

Gowiem:
@mikedizon can you rebase or pull in main? The pre-commit fix was merged so getting that upstream change into your branch should fix that CI check.

PR #409, opened 2022-04-28
Feat: Update `datadog-agent` component

Benbentwo:
## what
* Includes methodology for HTTP Cluster Checks

## why
* HTTP Checks are easy and lighter (cheaper too) than synthetics

## references
* N/A

### DIFF
--- a/modules/datadog-agent/default.auto.tfvars
+++ b/modules/datadog-agent/default.auto.tfvars
@@ -12,9 +12,7 @@ repository = "https://helm.datadoghq.com"
 
 chart = "datadog"
 
-# https://github.com/DataDog/helm-charts/releases
-# 2.26.2 released on Nov 19, 2021
-chart_version = "2.26.2"
+chart_version = "2.32.6"
 
 timeout = 180
 
--- a/modules/datadog-agent/helm-variables.tf
+++ b/modules/datadog-agent/helm-variables.tf
@@ -1,0 +1,63 @@
+variable "description" {
+  type        = string
+  description = "Release description attribute (visible in the history)"
+  default     = null
+}
+
+variable "chart" {
+  type        = string
+  description = "Chart name to be installed. The chart name can be local path, a URL to a chart, or the name of the chart if `repository` is specified. It is also possible to use the `<repository>/<chart>` format here if you are running Terraform on a system that the repository has been added to with `helm repo add` but this is not recommended"
+}
+
+variable "repository" {
+  type        = string
+  description = "Repository URL where to locate the requested chart"
+  default     = null
+}
+
+variable "chart_version" {
+  type        = string
+  description = "Specify the exact chart version to install. If this is not specified, the latest version is installed"
+  default     = null
+}
+
+variable "kubernetes_namespace" {
+  type        = string
+  description = "Kubernetes namespace to install the release into"
+}
+
+variable "timeout" {
+  type        = number
+  description = "Time in seconds to wait for any individual kubernetes operation (like Jobs for hooks). Defaults to `300` seconds"
+  default     = null
+}
+
+variable "cleanup_on_fail" {
+  type        = bool
+  description = "Allow deletion of new resources created in this upgrade when upgrade fails"
+  default     = true
+}
+
+variable "atomic" {
+  type        = bool
+  description = "If set, installation process purges chart on fail. The wait flag will be set automatically if atomic is used"
+  default     = true
+}
+
+variable "wait" {
+  type        = bool
+  description = "Will wait until all resources are in a ready state before marking the release as successful. It will wait for as long as `timeout`. Defaults to `true`"
+  default     = null
+}
+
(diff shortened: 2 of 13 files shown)


modules/datadog-agent/default.auto.tfvars
```diff
@@ -12,9 +12,7 @@ repository = "https://helm.datadoghq.com"
 
 chart = "datadog"
 
-# https://github.com/DataDog/helm-charts/releases
-# 2.26.2 released on Nov 19, 2021
```
nitrocode, line 16: If we keep this, then we can quickly see what the version is. Otherwise, you gotta google it, no ?

modules/datadog-agent/default.auto.tfvars
```diff
@@ -12,9 +12,7 @@ repository = "https://helm.datadoghq.com"
...
-chart_version = "2.26.2"
+chart_version = "2.32.6"
```
nitrocode, line 15: Also I think we're moving all of these out of default.auto.tfvars now, right ?

modules/datadog-agent/main.tf
```diff
@@ -14,24 +14,87 @@ locals {
...
-  repository       = var.repository
-  chart_version    = var.chart_version
-  namespace        = var.kubernetes_namespace
-  create_namespace = var.create_namespace
-  verify           = var.verify
-  wait             = var.wait
-  atomic           = var.atomic
-  cleanup_on_fail  = var.cleanup_on_fail
-  timeout          = var.timeout
+  version = "0.3.2"
```
nitrocode, line 84: Shouldn't this use the latest version ?

modules/datadog-agent/default.auto.tfvars
Benbentwo: I was thinking we remove that. Googling it can be a pain, but how often is that going to be the case when we can have a bot update this

modules/datadog-agent/default.auto.tfvars
Benbentwo: AFAIK we were keeping them in `default.auto.tfvars`

modules/datadog-agent/default.auto.tfvars
nitrocode: ya but when will that happen

nitrocode (approved):

modules/datadog-agent/providers.tf
```diff
@@ -58,10 +58,4 @@ data "aws_eks_cluster_auth" "kubernetes" {
...
 }
 
-provider "helm" {
-  kubernetes {
-    host                   = local.enabled ? data.aws_eks_cluster.kubernetes[0].endpoint : null
-    token                  = local.enabled ? data.aws_eks_cluster_auth.kubernetes[0].token : null
-    cluster_ca_certificate = local.enabled ? base64decode(data.aws_eks_cluster.kubernetes[0].certificate_authority[0].data) : null
-  }
-}
+provider "utils" {}
```
goruha, line 61: Do we use utils provider?

goruha (changes requested):

modules/datadog-agent/main.tf
```diff
@@ -14,24 +16,89 @@ locals {
...
+    for cluster_check_key, cluster_check_value in local.datadog_cluster_checks :
+    {
+      # Since we are using json pathing to set deep yaml values, and the key we want to set is `something.yaml`
+      # we need to escape the key of the cluster check.
+      name  = format("clusterAgent.confd.%s", replace(cluster_check_key, ".", "\\."))
+      type  = "auto"
+      value = yamlencode(cluster_check_value)
+    }
+  ]
+}
```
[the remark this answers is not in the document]
Benbentwo: We do use deep merging,  adding comments.

modules/datadog-agent/providers.tf
Benbentwo: Yes https://github.com/cloudposse/terraform-aws-components/pull/409/files#diff-9d977686018b77e00be742c079a4f3c3259a1d9266f8dc0c8036837064b10e78R57 uses it as part of the module now

modules/datadog-agent/helm-variables.tf
```diff
@@ -0,0 +1,63 @@
...
+
+variable "atomic" {
+  type        = bool
+  description = "If set, installation process purges chart on fail. The wait flag will be set automatically if atomic is used"
+  default     = true
+}
+
+variable "wait" {
+  type        = bool
+  description = "Will wait until all resources are in a ready state before marking the release as successful. It will wait for as long as `timeout`. Defaults to `true`"
```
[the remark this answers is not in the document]
Benbentwo: This is pulled from the helm provider https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#wait

we default to null to use that provider default

nitrocode (approved):

PR #410, opened 2022-05-04
Add eks (current flavor)

nitrocode:
## what
* [x] Add eks (current flavor)
* [x] Rename and update old eks component as `eks-without-spotinst`

## why
* This will be possible thanks to the recent atmos feature to allow vendoring
* The plan is to have different flavors for our components e.g. `eks-without-spotinst`, `eks-with-spotinst`
* Requested by @osterman 

## references
* Closes https://github.com/cloudposse/terraform-aws-components/pull/397
* Closes https://github.com/cloudposse/terraform-aws-components/pull/393
* Closes https://github.com/cloudposse/terraform-aws-components/pull/373


modules/eks/eks-without-spotinst/variables.tf
```diff
@@ -0,0 +1,346 @@
...
+
+variable "cluster_kubernetes_version" {
+  type        = string
+  default     = null
+  description = "Desired Kubernetes master version. If you do not specify a value, the latest available version is used"
+}
+
+variable "public_access_cidrs" {
+  type        = list(string)
+  default     = ["0.0.0.0/0"]
```
mcalhoun, line 52: I know this is probably "how it is currently", but this seems like a bad default. We should probably default to the more secure option of not allowing any access and adding on as needed.

modules/eks/eks-without-spotinst/variables.tf
```diff
@@ -0,0 +1,346 @@
...
+  type        = list(string)
+  default     = []
+  description = "A list of the desired control plane logging to enable. For more information, see https://docs.aws.amazon.com/en_us/eks/latest/userguide/control-plane-logs.html. Possible values [`api`, `audit`, `authenticator`, `controllerManager`, `scheduler`]"
+}
+
+variable "cluster_log_retention_period" {
+  type        = number
+  default     = 0
+  description = "Number of days to retain cluster logs. Requires `enabled_cluster_log_types` to be set. See https://docs.aws.amazon.com/en_us/eks/latest/userguide/control-plane-logs.html."
+}
```
mcalhoun, line 66: Kind of the same comment as above. We should probably default to a best practice and let customers opt out as needed.

modules/eks/eks-without-spotinst/variables.tf
```diff
@@ -0,0 +1,346 @@
...
+
+variable "cluster_private_subnets_only" {
+  type        = bool
+  default     = false
+  description = "Whether or not to enable private subnets or both public and private subnets"
+}
+
+variable "allow_ingress_from_vpc_stages" {
+  type        = list(string)
+  default     = ["auto", "corp"]
```
mcalhoun, line 338: IMO defaulting to no access and having this be part of the catalog would be a better choice

mcalhoun (approved):
I left a few small comments that are probably from the original implementation, but maybe now is the time to address them? Otherwise, looks good to me.

modules/eks/eks-without-spotinst/variables.tf
```diff
@@ -0,0 +1,346 @@
...
+
+variable "enabled_cluster_log_types" {
+  type        = list(string)
+  default     = []
+  description = "A list of the desired control plane logging to enable. For more information, see https://docs.aws.amazon.com/en_us/eks/latest/userguide/control-plane-logs.html. Possible values [`api`, `audit`, `authenticator`, `controllerManager`, `scheduler`]"
+}
+
+variable "cluster_log_retention_period" {
+  type        = number
+  default     = 0
```
nitrocode, line 64:
```suggestion
  default     = 90
```

modules/eks/eks-without-spotinst/variables.tf
nitrocode: Defaulted to 90

modules/eks/eks-without-spotinst/variables.tf
nitrocode: Defaulted to []

mcalhoun (approved):
LGTM!

PR #412, opened 2022-05-17
Upstream `s3-bucket`

milldr:
## what
- Upstreaming `s3-bucket`

## why
- Using the latest versions and changes for `s3-bucket` component

## references
- Required to set up vendoring for CPLIVE, https://github.com/cloudposse/infra-live/pull/116

## NOTE
* If you are adding a new component to `terraform-aws-components`, you must add a label to the pull request specifying the earliest major and minor versions of Terraform for which the component must remain backwards compatible. (E.g., `terraform/0.15`, `terraform/0.12`, etc.). If compatibility with pre-1.0 versions of Terraform is not required, just select the `terraform/1.x` label.
* The chosen label will determine the version of Terraform that will be used in this pull request's `pre-commit` status checks.


modules/s3-bucket/main.tf
```diff
@@ -1,53 +1,83 @@
...
     for acct in var.custom_policy_account_names :
-    format("arn:%s:iam::%s:root", data.aws_partition.current.partition, module.account_map.outputs.full_account_map[acct])
+    format("arn:%s:iam::%s:root", local.aws_partition, module.account_map.outputs.full_account_map[acct])
   ]
 
-  bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : var.policy
+  bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : data.template_file.bucket_policy.rendered
 }
 
 data "aws_partition" "current" {}
```
Nuru, line 14: This should not be necessary. If we are requiring `account_map` anyway, we should use `account_map/modules/roles_to_principals` to map from account names to principal Role ARNs. (`roles_to_principals` should take `partition` as a variable with default `aws`.)

modules/s3-bucket/main.tf
```diff
@@ -1,53 +1,83 @@
...
-  restrict_public_buckets                = var.restrict_public_buckets
-  s3_replica_bucket_arn                  = var.s3_replica_bucket_arn
-  s3_replication_enabled                 = var.s3_replication_enabled
-  s3_replication_rules                   = var.s3_replication_rules
-  s3_replication_source_roles            = var.s3_replication_source_roles
-  sse_algorithm                          = var.sse_algorithm
-  transfer_acceleration_enabled          = var.transfer_acceleration_enabled
-  versioning_enabled                     = var.versioning_enabled
-  website_inputs                         = var.website_inputs
+  version = "2.0.0"
```
Nuru, line 35: Release notes need to reference [s3-bucket v2.0 upgrade instructions](https://github.com/cloudposse/terraform-aws-s3-bucket/wiki/Upgrading-to-v2.0) and in particular warn that S3-backed static websites are broken.

modules/s3-bucket/main.tf
```diff
@@ -1,53 +1,83 @@
...
+  ignore_public_acls           = var.ignore_public_acls
+  restrict_public_buckets      = var.restrict_public_buckets
+  logging                      = var.logging
+  source_policy_documents      = [local.bucket_policy]
+  privileged_principal_actions = var.privileged_principal_actions
+  privileged_principal_arns    = var.privileged_principal_arns
+
+  # Static website configuration
+  cors_rule_inputs = var.cors_rule_inputs
+  website_inputs   = var.website_inputs
```
Nuru, line 55: Note: `website_inputs` does not work in `s3-buckets` v2.0.0 (see release notes).

modules/s3-bucket/main.tf
```diff
@@ -1,53 +1,83 @@
...
+  s3_replica_bucket_arn       = var.s3_replica_bucket_arn
+  s3_replication_rules        = var.s3_replication_rules
+  s3_replication_source_roles = var.s3_replication_source_roles
 
   # IAM user with permissions to access the s3 bucket
   user_enabled           = var.user_enabled
   allowed_bucket_actions = var.allowed_bucket_actions
 
-  context = module.this.context
+  context = module.introspection.context
```
Nuru, line 80: Let's not use `introspection` in `terraform-aws-components`, at least not yet.

modules/s3-bucket/main.tf
```diff
@@ -57,11 +87,11 @@ data "aws_iam_policy_document" "custom_policy" {
     actions = var.custom_policy_actions
 
     resources = [
-      format("arn:aws:s3:::%s-%s-%s-%s", module.this.namespace, module.this.environment, module.this.stage, module.this.name),
-      format("arn:aws:s3:::%s-%s-%s-%s/*", module.this.namespace, module.this.environment, module.this.stage, module.this.name)
+      format("arn:%s:s3:::%s", local.aws_partition, module.this.id),
+      format("arn:%s:s3:::%s/*", local.aws_partition, module.this.id)
```
Nuru, line 91: We should not be constructing ARNs like this. Use
```suggestion
      module.s3_bucket.bucket_id,
      "${module.s3_bucket.bucket_id}/*",
```

modules/s3-bucket/main.tf
```diff
@@ -57,11 +87,11 @@ data "aws_iam_policy_document" "custom_policy" {
...
     ]
     principals {
-      identifiers = local.custom_policy_account_arns
+      identifiers = length(local.custom_policy_account_arns) > 0 ? local.custom_policy_account_arns : ["*"]
```
Nuru, line 94: I do not like the default being fully open access. That is dangerous. Default should be an error, because it is easy enough to specify `["*"]` if that is what you want.
```suggestion
      identifiers = local.custom_policy_account_arns
```

modules/s3-bucket/providers.tf
```diff
@@ -11,8 +12,9 @@ provider "aws" {
 }
 
 module "iam_roles" {
-  source  = "../account-map/modules/iam-roles"
-  context = module.this.context
+  source             = "../account-map/modules/iam-roles"
+  global_tenant_name = var.root_account_tenant_name
```
Nuru, line 16: This should not be needed. `iam-roles` should be configured with this information specifically so we do not need to pass it through all the root modules.
```suggestion
```

modules/s3-bucket/providers.tf
```diff
@@ -26,3 +28,9 @@ variable "import_role_arn" {
   default     = null
   description = "IAM Role ARN to use when importing a resource"
 }
+
+variable "root_account_tenant_name" {
+  type        = string
+  description = "The tenant name for the root account"
+  default     = null
+}
```
Nuru, line 36: This should not be needed. `iam-roles` should be configured with this information specifically so we do not need to pass it through all the root modules.
```suggestion
```

modules/s3-bucket/remote-state.tf
```diff
@@ -1,6 +1,6 @@
 module "account_map" {
   source  = "cloudposse/stack-config/yaml//modules/remote-state"
-  version = "0.22.0"
+  version = "0.22.1"
```
Nuru, line 3:
```suggestion
  version = "0.22.2"
```

modules/s3-bucket/variables.tf
```diff
@@ -106,50 +119,59 @@ variable "allow_ssl_requests_only" {
...
-    noncurrent_version_expiration_days             = 90
-
-    standard_transition_days    = 30
-    glacier_transition_days     = 60
-    deeparchive_transition_days = 90
-    expiration_days             = 90
-  }]
-  description = "A list of lifecycle rules"
+  default     = []
+  description = "A list of lifecycle V2 rules"
```
Nuru, line 174:
```suggestion
  description = "A list of lifecycle V2 rules. See [terraform-aws-s3-bucket](https://github.com/cloudposse/terraform-aws-s3-bucket) for details."
```

modules/s3-bucket/variables.tf
```diff
@@ -228,6 +245,9 @@ variable "s3_replication_rules" {
   #       owner = string
   #     })
   #     account_id                 = string
+  #     metrics                    = object({
+  #       status = string
+  #     })
```
Nuru, line 250: Add "See [terraform-aws-s3-bucket](https://github.com/cloudposse/terraform-aws-s3-bucket) for details." to variable description.

modules/s3-bucket/versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 3.0"
+      version = "~> 4.0"
```
Nuru, line 7:
```suggestion
      version = "~> 4.9"
```

Nuru (changes requested):

modules/s3-bucket/main.tf
milldr: removed

modules/s3-bucket/providers.tf
milldr: To configure `iam-roles` with the root account tenant name, do we need to set the default value for `var.global_tenant_name` within `.../account-map/modules/iam-roles/variables.tf`? 
```
variable "global_tenant_name" {
  type        = string
  description = "Global tenant name"
  default     = "core" # this line
}
```
That does work, but I thought we would want to keep all variables values within components as non-specific. Is there another way to set that value that I'm missing?

modules/s3-bucket/main.tf
```diff
@@ -1,17 +1,26 @@
...
+  bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : data.template_file.bucket_policy.rendered
+}
+
+module "custom_policy_account_roles" {
+  count = local.enabled && var.custom_policy_enabled ? 1 : 0
+
+  source = "../account-map/modules/roles-to-principals"
 
-  custom_policy_account_arns = [
+  iam_role_arn_template = "arn:aws:iam::%s:role/%s-%s-%s-%s"
```
Nuru, line 12: We should not need this. That value should be pre-configured in `account-map`.
```suggestion
```

modules/s3-bucket/main.tf
```diff
@@ -1,17 +1,26 @@
...
-    format("arn:%s:iam::%s:root", local.aws_partition, module.account_map.outputs.full_account_map[acct])
-  ]
+    acct => ["root"]
+  }
 
-  bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : data.template_file.bucket_policy.rendered
-}
+  root_account_stage_name = var.account_map_stage_name
+  global_environment_name = var.account_map_environment_name
+  global_tenant_name      = coalesce(var.account_map_tenant_name, module.this.tenant)
```
Nuru, line 20: We should not need these. Those values should be pre-configured in `account-map`.
```suggestion
```

modules/s3-bucket/main.tf
```diff
@@ -27,10 +36,15 @@ module "bucket_policy" {
...
-  context = module.introspection.context
+  context = module.this.context
 }
 
 module "s3_bucket" {
+  # Note: 
+  # Version 2.0 introduces a breaking change to var.website_inputs. 
+  # The AWS provider v3 configuration of S3 static websites was replaced with a new aws_s3_bucket_website_configuration resource which has additional features. The s3-bucket v1 website_inputs is broken (due to an oversight) in v2 and is inadequate to handle the new features.
+  # See the 2.0 release notes:
+  # https://github.com/cloudposse/terraform-aws-s3-bucket/wiki/Upgrading-to-v2.0
```
Nuru, line 47: This also needs to go in the PR description and eventual release notes.

modules/s3-bucket/main.tf
```diff
@@ -52,7 +66,7 @@ module "s3_bucket" {
 
   # Static website configuration
   cors_rule_inputs = var.cors_rule_inputs
-  website_inputs   = var.website_inputs
+  # website_inputs   = var.website_inputs
```
Nuru, line 69: Consider taking over/finishing https://github.com/cloudposse/terraform-aws-s3-bucket/pull/142 so we can have this module support websites. Otherwise, this lack of support needs to be prominently documented.

modules/s3-bucket/main.tf
```diff
@@ -87,11 +101,11 @@ data "aws_iam_policy_document" "custom_policy" {
...
 
     resources = [
-      format("arn:%s:s3:::%s", local.aws_partition, module.this.id),
-      format("arn:%s:s3:::%s/*", local.aws_partition, module.this.id)
+      module.s3_bucket.bucket_arn,
+      "${module.s3_bucket.bucket_arn}/*",
     ]
     principals {
-      identifiers = length(local.custom_policy_account_arns) > 0 ? local.custom_policy_account_arns : ["*"]
+      identifiers = module.custom_policy_account_roles.principals
```
Nuru, line 108: We probably want to preserve and option of creating an open bucket.

Nuru (changes requested):

milldr:
We are putting this upstream on hold. The required upstream for CPLIVE-266 will be included in PR #417

milldr:
this upstream is outdated now

PR #439, opened 2022-07-21
Adding new `sns-topic` module

danjbh:
## what
* Adding new `sns-topic` module

## why
* Because we don't have one yet!


modules/sns-topic/default.auto.tfvars
```diff
@@ -0,0 +1,5 @@
+# This file is included by default in terraform plans
+enabled = false
+
+name = "sns-topic"
+
```
nitrocode, line 5: all of this we can keep in the catalog except for enabled=false

```suggestion
```

modules/sns-topic/README.md
```diff
@@ -0,0 +1,153 @@
...
+**Stack Level**: Regional
+
+Here are some example snippets for how to use this component:
+
+`stacks/catalog/sns-topic/defaults.yaml` file (base component for all SNS topics with default settings):
+
+```yaml
+components:
+  terraform:
+    sns-topic-defaults:
```
nitrocode, line 16:
```suggestion
    sns-topic/defaults:
```

modules/sns-topic/README.md
```diff
@@ -0,0 +1,153 @@
...
+import:
+- catalog/sns-topic/defaults
+
+components:
+  terraform:
+    template-bucket:
+      metadata:
+        component: sns-topic
+        inherits:
+          - sns-topic-defaults
```
nitrocode, line 56:
```suggestion
          - sns-topic/defaults
```

nitrocode (commented):
Nice addition. Please see the comments and then LGTM

modules/sns-topic/default.auto.tfvars
```diff
@@ -0,0 +1,5 @@
+# This file is included by default in terraform plans
+enabled = false
+
+name = "sns-topic"
```
milldr, line 4:
```suggestion
```

modules/sns-topic/README.md
danjbh: hey @nitrocode, I was following the latest pattern and using `s3-bucket` as an example, which uses something similar to what I have -- https://github.com/cloudposse/terraform-aws-components/tree/master/modules/s3-bucket

Have a look and let me know what you think.

nitrocode (dismissed):

modules/sns-topic/README.md
nitrocode: I prefer the forward slash since this creates the nice folder structure in spacelift :)

nitrocode: It's also more intuitive IMHO because `sns-topic/defaults` matches the catalog file path

danjbh: Fair enough.. I'll make the change.

nitrocode (approved):

PR #442, opened 2022-07-29
add ignore_changes block for iam_user_access_to_billing to account resources

joe-niland:
## what
* Ignore changes to the iam_user_access_to_billing attribute on `aws_organizations_account.organization_accounts` and `aws_organizations_account.organizational_units_accounts`
* This is mainly to support importing of accounts, but also would stop the account being replaced if the value is changed, which is probably what is wanted in most cases.

## why
* The AWS [API](https://docs.aws.amazon.com/organizations/latest/APIReference/API_DescribeAccount.html) doesn't allow this value to be read
* When an account is imported, `iam_user_access_to_billing` isn't imported due to the value not existing in the API response. Upon running the account component again, TF wants to destroy the resource to apply this value.
* Other than editing the TF state, this lifecycle block will resolve the issue.

## references
* None


nitrocode (approved):

Gowiem:
Thanks for the assist @nitrocode!

PR #444, opened 2022-08-01
Adding new 'lakeformation' component

danjbh:
## what & why?
* New 'lakeformation' component to deploy Amazon Lake formation resources in AWS, w/ docs and sample config


milldr (approved):
LGTM

PR #455, opened 2022-08-12
upstream `alb-controller`

Benbentwo:
## what
* Upstream ALB Controller


modules/alb-controller/default.auto.tfvars
```diff
@@ -1,12 +1,15 @@
+# This file is included by default in terraform plans
+
 enabled = false
 
 name = "alb-controller"
```
nitrocode, line 5: All of these default auto tfvars should be removed except for `enabled`

modules/alb-controller/main.tf
```diff
@@ -2,23 +2,35 @@ locals {
...
+  version = "0.5.0"
 
   name                 = "" # avoids hitting length restrictions on IAM Role names
   chart                = var.chart
   repository           = var.chart_repository
   description          = var.chart_description
   chart_version        = var.chart_version
-  kubernetes_namespace = var.kubernetes_namespace
-  create_namespace     = var.create_namespace
+  kubernetes_namespace = join("", kubernetes_namespace.default.*.id)
```
nitrocode, line 26: This line above assumes that the `var.create_namespace` will always be true.

We should solve for the case when it is false by passing in the `kubernetes_namespace` input directly

```suggestion
  kubernetes_namespace = local.enabled && var.create_namespace ? join("", kubernetes_namespace.default.*.id) : var.kubernetes_namespace
```

modules/alb-controller/providers.tf
```diff
@@ -1,11 +1,18 @@
...
+    for_each = module.iam_roles.profiles_enabled ? [] : ["role"]
+    content {
+      role_arn = coalesce(var.import_role_arn, module.iam_roles.terraform_role_arn)
+    }
+  }
 }
 
 module "iam_roles" {
-  source  = "../account-map/modules/iam-roles"
+  source  = "../../account-map/modules/iam-roles"
```
nitrocode, line 15: We should probably move this module into the `eks` directory otherwise this path doesn't make sense

nitrocode: But we don't have a precedence for this yet so I'm ok with keeping it where it is for now

nitrocode (approved):

modules/github-runners/README.md
```diff
@@ -146,7 +146,7 @@ To generate the token, please follow [these instructions](https://cloudposse.atl
 | <a name="input_region"></a> [region](#input\_region) | AWS Region | `string` | n/a | yes |
 | <a name="input_runner_labels"></a> [runner\_labels](#input\_runner\_labels) | List of labels to add to the GitHub Runner (e.g. 'Amazon Linux 2'). | `list(string)` | `[]` | no |
 | <a name="input_runner_role_additional_policy_arns"></a> [runner\_role\_additional\_policy\_arns](#input\_runner\_role\_additional\_policy\_arns) | List of policy ARNs that will be attached to the runners' default role on creation in addition to the defaults | `list(string)` | `[]` | no |
-| <a name="input_runner_version"></a> [runner\_version](#input\_runner\_version) | GitHub runner release version | `string` | `"2.287.1"` | no |
+| <a name="input_runner_version"></a> [runner\_version](#input\_runner\_version) | GitHub runner release version | `string` | `"2.288.1"` | no |
```
milldr, line 149: remove this file from this PR

milldr (approved):

PR #460, opened 2022-08-18
Adding new 'mwaa' component

danjbh:
## what
* Adding new `mwaa` component for Amazon Airflow deployments

### DIFF
--- a/modules/mwaa/iam.tf
+++ b/modules/mwaa/iam.tf
@@ -1,6 +1,6 @@
 module "iam_policy" {
   source  = "cloudposse/iam-policy/aws"
-  version = "0.3.0"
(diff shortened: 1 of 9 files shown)


modules/mwaa/iam.tf
```diff
@@ -0,0 +1,31 @@
+module "iam_policy" {
+  source  = "cloudposse/iam-policy/aws"
+  version = "0.3.0"
```
Benbentwo, line 3: newer version is out
```suggestion
  version = "0.4.0"
```

Benbentwo (changes requested):

Benbentwo (approved):

PR #469, opened 2022-08-23
Set default terraform version rather than requiring (inconsistent) tags

Nuru:
## what
- Use a default Terraform version for formatting when no labels are present

## why
- Requiring a label per PR to determine the Terraform version is extra work that also leads to likely "flapping" if one PR uses a different label than another. Labels should only be required when there is some reason not to use the default.

### DIFF
--- a/.github/workflows/pre-commit-check-and-autocommit-changes.yaml
+++ b/.github/workflows/pre-commit-check-and-autocommit-changes.yaml
@@ -42,6 +42,7 @@ jobs:
           BASE_REF: ${{ github.base_ref }}
           LABELS: ${{ join(github.event.pull_request.labels.*.name, '\n') }}
           MODIFIED_FILES: ${{ steps.get-modified-files.outputs.modified_files }}
+          DEFAULT_TERRAFORM_VERSION: ${{ secrets.DEFAULT_TERRAFORM_VERSION }}
         run: |
           # Match labels like `terraform/0.12` or nothing (to prevent grep from returning a non-zero exit code)
           # Use [0-9] because \d is not standard part of egrep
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit-check-and-autocommit-changes.yaml
```diff
@@ -51,17 +52,8 @@ jobs:
           if grep -Ez '^\W*[0-9]+\.[x0-9]+\W*$' <<<${TERRAFORM_VERSION}; then
             echo "Terraform version ${TERRAFORM_VERSION} will be used to check the formatting of files that have been modified since this branch diverged from ${BASE_REF}."
           elif [ -z "${TERRAFORM_VERSION}" ]; then
-            # If there are no compliant labels, it could be that the PR doesn't need Terraform linting in the first place
-            # To detemine this, check to see if any ".tf" files are modified.
-            if grep -E '.tf\>' <<<${MODIFIED_FILES}; then
-              echo "Terraform files have been modified on this branch, but there is no label on the pull request indicating which version of Terraform they are meant to be compliant with."
-              echo "A label is required indicating the version of Terraform to be used for formatting this component. (E.g., \"terraform/1.x\" or \"terraform/0.15\")"
-              exit 1
-            else
```
dylanbannon, line 60: My only concern with this PR is that we now have a situation where someone could modify a component that's meant to be pre-`1.x` compliant (does we still have those?), then they open the PR and forget to assign the `0.15` or whatever label, see that the checks pass (even though the checks that ran were the `1.x` defaults), merge, and then end up making their component incompatible with TF `0.15`, etc. Is that possible?

osterman (approved):

PR #471, opened 2022-08-24
`cognito` component

Benbentwo:
## what
* AWS Cognito Component

## why
* Upstream

### DIFF
--- a/modules/cognito/versions.tf
+++ b/modules/cognito/versions.tf
@@ -6,7 +6,7 @@ terraform {
       source = "hashicorp/aws"
       # Starting with v4.8.0, the provider adds `aws_cognito_user_in_group` allowing adding Cognito Users to Cognito Groups in terraform
       # https://github.com/hashicorp/terraform-provider-aws/releases
-      version = "~> 4.8.0"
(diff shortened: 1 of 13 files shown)


modules/cognito/versions.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_version = ">= 1.0.0"
+
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      # Starting with v4.8.0, the provider adds `aws_cognito_user_in_group` allowing adding Cognito Users to Cognito Groups in terraform
+      # https://github.com/hashicorp/terraform-provider-aws/releases
+      version = "~> 4.8.0"
```
mcalhoun, line 9:
```suggestion
      version = ">= 4.8.0"
```

This way we don't have to update the module when we want it to use `4.9`, `4.10`, or even `5.0`

mcalhoun (changes requested):
Overall, this looks great! Just left a small, but IMO important, comment.

mcalhoun (approved):
LGTM!

PR #472, opened 2022-08-25
upstream `kms`

Benbentwo:
## what
* Simple KMS Component

## why
* Upstream

### DIFF
--- a/modules/kms/README.md
+++ b/modules/kms/README.md
@@ -1,6 +1,6 @@
 # Component: `kms`
 
-This component is responsible for creating an kms Key.
+This component is responsible for provisioning a KMS Key.
 
 ## Usage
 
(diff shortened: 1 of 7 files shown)


modules/kms/providers.tf
```diff
@@ -0,0 +1,29 @@
+module "iam_roles" {
```
milldr, line 1: nitpick: the content is the same, but the modules are in a different order than usual in this file. Could we replace this file with the same, latest provider.tf that we've been using in other components? Might help prevent future confusion

modules/kms/README.md
```diff
@@ -0,0 +1,91 @@
+# Component: `kms`
```
milldr, line 1: add `versions.tf`

milldr (approved):

PR #473, opened 2022-08-29
Upstream Aurora MySQL components

milldr:
## what
- Updating the `aurora-mysql` and `aurora-mysql-resources` components

## why
- Add support for cross region Aurora Read Replicas
- Correct issues with SSM Parameters in both components
- Further debugging, clean up, and TLC for the components


modules/aurora-mysql-resources/variables.tf
```diff
@@ -17,7 +23,7 @@ variable "ssm_path_prefix" {
 variable "ssm_password_source" {
   type        = string
   default     = ""
-  description = "If set, DB user passwords will be retrieved from SSM using the key `format(var.ssm_password_source, local.db_username)`"
+  description = "If var.ssm_passwords_enabled is true, DB user passwords will be retrieved from SSM using the key `format(var.ssm_password_source, local.db_username)`. If this value is not set, a default path will be created using the SSM path prefix and ID of the associated Aurora Cluster."
```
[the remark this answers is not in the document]
milldr: updated description

modules/aurora-mysql/outputs.tf
```diff
@@ -1,51 +1,62 @@
...
+  description = "The ID of Aurora Cluster"
 }
 
-output "aurora_mysql_master_password" {
-  value       = local.mysql_enabled ? "Password for admin user ${module.aurora_mysql.master_username} is stored in SSM at ${local.mysql_admin_password_key}" : null
-  description = "Location of admin password"
-  sensitive   = true
+output "aurora_mysql_cluster_name" {
+  value       = local.enabled ? module.aurora_mysql.cluster_identifier : null
+  description = "RDS Aurora-MySQL: Cluster Identifier"
```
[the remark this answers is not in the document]
milldr: replaced all: `%s/RDS Aurora-MySQL:/Aurora MySQL/g`

modules/aurora-mysql-resources/variables.tf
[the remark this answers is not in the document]
milldr: updated description for similar variable in `aurora-mysql` as well

modules/aurora-mysql/variables.tf
```diff
@@ -12,7 +12,7 @@ variable "ssm_path_prefix" {
 variable "ssm_password_source" {
   type        = string
   default     = ""
-  description = "If set, DB Admin user password will be retrieved from SSM using the key `format(var.ssm_password_source, local.db_username)`"
+  description = "If var.ssm_passwords_enabled is true, DB user passwords will be retrieved from SSM using `var.ssm_password_source` and the database username. If this value is not set, a default path will be created using the SSM path prefix and ID of the associated Aurora Cluster."
```
Nuru, line 15:
```suggestion
  description = <<-EOT
    If `var.ssm_passwords_enabled` is `true`, DB user passwords will be retrieved from SSM using
    `var.ssm_password_source` and the database username. If this value is not set, a default path
    will be created using the SSM path prefix and ID of the associated Aurora Cluster.
    EOT
```

modules/aurora-mysql/variables.tf
```diff
@@ -159,8 +153,33 @@ variable "publicly_accessible" {
...
+  type        = set(string)
+  description = "The names of the eks components"
+  default     = ["eks/cluster"]
+}
+
+variable "replication_source_identifier" {
   type        = string
-  description = "The name of the eks component"
-  default     = "eks/eks"
+  description = "ARN of a source DB cluster or DB instance if this DB cluster is to be created as a Read Replica. If this value is empty and replication is enabled, remote state will attempt to find a matching cluster in the Primary DB Cluster's region"
```
Nuru, line 164:
```suggestion
  description = <<-EOT
    ARN of a source DB cluster or DB instance if this DB cluster is to be created as a Read Replica.
    If this value is empty and replication is enabled, remote state will attempt to find a matching 
    cluster in the Primary DB Cluster's region.
    EOT
```

modules/aurora-mysql/variables.tf
```diff
@@ -159,8 +153,33 @@ variable "publicly_accessible" {
...
+  default     = ""
 }
+
+variable "replication_enabled" {
+  type        = bool
+  description = "Whether or not to create this DB cluster as a Read Replica."
```
Nuru, line 170: This is confusing. I would expect I would need to set `replication_enabled` on the primary cluster in order to create a read replica. I think if this is truly a flag on whether the component is itself a read replica or not, it should be called `is_read_replica` or something like that. We also need to be able to reconfigure these components to promote the read replica to primary and vice versa for disaster recovery, so we might want to call this `is_primary`. @aknysh @osterman What do you think?

Nuru (changes requested):

modules/aurora-mysql-resources/main.tf
```diff
@@ -47,15 +51,15 @@ module "additional_grants" {
...
-  save_password_in_ssm = local.fetch_passwords ? false : true
-  kms_key_id           = local.kms_key_arn
+  service_name = each.key
+  grants       = each.value
+  kms_key_id   = local.kms_key_arn
+
+  # If no password is given, a random password will be created
+  db_password = local.ssm_passwords_enabled ? data.aws_ssm_parameter.password[each.key].value : ""
+  # If generating a password, store it in SSM
+  save_password_in_ssm = local.ssm_passwords_enabled ? false : true
```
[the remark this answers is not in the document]
milldr: `ssm_passwords_enabled` means passwords are stored in SSM, so we don't want to generate a new password and save that password in SSM. Added clarification to that comment

modules/aurora-mysql-resources/modules/mysql-user/main.tf
```diff
@@ -1,15 +1,17 @@
...
-  db_user         = length(var.db_user) > 0 ? var.db_user : var.service_name
-  db_password     = length(var.db_password) > 0 ? var.db_password : join("", random_password.db_password.*.result)
-  db_password_key = format("/%s/%s/%s", var.ssm_path_prefix, var.service_name, "db_password")
+  db_user     = length(var.db_user) > 0 ? var.db_user : var.service_name
+  db_password = length(var.db_password) > 0 ? var.db_password : join("", random_password.db_password.*.result)
 
   create_db_user       = local.enabled && var.service_name != local.db_user
   save_password_in_ssm = local.enabled && var.save_password_in_ssm
 
+  db_user_key     = format("%s/%s/%s", var.ssm_path_prefix, var.service_name, "db_user")
```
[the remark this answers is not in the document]
milldr: updated the keys for both username and password to include the user name:
```
  db_user_key     = format("%s/%s/users/%s/username", var.ssm_path_prefix, var.service_name, local.db_user)
  db_password_key = format("%s/%s/users/%s/password", var.ssm_path_prefix, var.service_name, local.db_user)
```

modules/aurora-mysql-resources/variables.tf
```diff
@@ -8,6 +8,12 @@ variable "aurora_mysql_component_name" {
   description = "Aurora MySQL component name to read the remote state from"
 }
 
+variable "ssm_passwords_enabled" {
+  type        = bool
+  default     = true
+  description = "Whether or not to use SSM for user passwords"
```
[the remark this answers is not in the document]
milldr: `"When `true`, fetch user passwords from SSM"`

modules/aurora-mysql-resources/variables.tf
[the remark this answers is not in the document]
milldr: updated with re-worded description

modules/aurora-mysql/kms.tf
```diff
@@ -59,7 +57,8 @@ data "aws_iam_policy_document" "kms_key_rds" {
       "kms:Decrypt*",
       "kms:ReEncrypt*",
       "kms:GenerateDataKey*",
-      "kms:Describe*"
+      "kms:Describe*",
+      "kms:CreateGrant"
```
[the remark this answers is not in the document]
milldr: Yes it is required for creating the replica. Terraform errored out without it, and this [SO thread ](https://stackoverflow.com/questions/45821144/minimal-kms-permissions-to-copy-a-database-snapshot)confirmed it

modules/aurora-mysql/variables.tf
milldr: fair enough, I will update the variable name. But Replication should be enabled by the replica component, since it's deployed independently from the primary cluster, with a separate task/ticket/stack, and in a different region. The replica can be destroyed without worrying about affecting the primary cluster

milldr: this is for cross-region replication specifically. We need to create / update docs for the differences between global, single regional, cross regional

milldr: updated variable name and related descriptions

modules/aurora-mysql-resources/main.tf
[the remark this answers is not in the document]
Nuru: This `ssm_passwords_enabled` remains confusing to me. We have similar variables in our open source modules, and they man "save passwords (whether configured or generated) in SSM". I suggest you rename this to `read_passwords_from_ssm`.

modules/aurora-mysql-resources/modules/mysql-user/main.tf
```diff
@@ -7,8 +7,8 @@ locals {
   create_db_user       = local.enabled && var.service_name != local.db_user
   save_password_in_ssm = local.enabled && var.save_password_in_ssm
 
-  db_user_key     = format("%s/%s/%s", var.ssm_path_prefix, var.service_name, "db_user")
-  db_password_key = format("%s/%s/%s", var.ssm_path_prefix, var.service_name, "db_password")
+  db_user_key     = format("%s/%s/users/%s/username", var.ssm_path_prefix, var.service_name, local.db_user)
```
Nuru, line 10: This seems needlessly redundant. is `users/foo/username` ever not going to be "foo"?

modules/aurora-mysql-resources/modules/mysql-user/main.tf
```diff
@@ -7,8 +7,8 @@ locals {
...
+  db_password_key = format("%s/%s/users/%s/password", var.ssm_path_prefix, var.service_name, local.db_user)
```
Nuru, line 11:
```suggestion
  db_password_key = format("%s/%s/passwords/%s", var.ssm_path_prefix, var.service_name, local.db_user)
```

modules/aurora-mysql-resources/modules/mysql-user/main.tf
```diff
@@ -91,7 +91,6 @@ resource "mysql_user" "default" {
   user               = local.db_user
   host               = "%"
   plaintext_password = local.db_password
-  #  depends_on         = [var.instance_ids]
```
Nuru, line 92: Are you sure this is not needed? What is preventing Terraform from trying to create the user before the database is created?

modules/aurora-mysql-resources/variables.tf
```diff
@@ -11,7 +11,7 @@ variable "aurora_mysql_component_name" {
 variable "ssm_passwords_enabled" {
```
Nuru, line 11:
```suggestion
variable "read_passwords_from_ssm" {
```
Like I said above, we have several other instances of variables with this or very similar names, and they all enable storing passwords in SSM. Let's use a more specifically descriptive name to indicate the different usage.

modules/aurora-mysql/main.tf
```diff
@@ -11,17 +11,18 @@ locals {
...
+  is_read_replica               = local.enabled && var.is_read_replica
+  remote_read_replica_enabled   = local.is_read_replica && !(length(var.replication_source_identifier) > 0) && length(var.primary_cluster_region) > 0
   replication_source_identifier = local.remote_read_replica_enabled ? module.primary_cluster[0].outputs.aurora_mysql_cluster_arn : var.replication_source_identifier
 
   # For encrypted cross-region replica, kmsKeyId should be explicitly specified
   # https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html
-  # A read replica of an Amazon RDS encrypted instance must be encrypted using the same KMS key as the primary DB instance when both are in the same AWS Region. If the primary DB instance and read replica are in different AWS Regions, you encrypt the read replica using the KMS key for that AWS Region.
+  # A read replica of an Amazon RDS encrypted instance must be encrypted using the same KMS key as the primary DB instance when both are in the same AWS Region. 
+  # If the primary DB instance and read replica are in different AWS Regions, you encrypt the read replica using the KMS key for that AWS Region.
```
Nuru, line 21:
```suggestion
  # If the primary DB instance and read replica are in different AWS Regions, you encrypt the read replica using a KMS key in that AWS Region.
```

modules/aurora-mysql/variables.tf
```diff
@@ -12,7 +12,9 @@ variable "ssm_path_prefix" {
 variable "ssm_password_source" {
   type        = string
   default     = ""
-  description = "If var.ssm_passwords_enabled is true, DB user passwords will be retrieved from SSM using `var.ssm_password_source` and the database username. If this value is not set, a default path will be created using the SSM path prefix and ID of the associated Aurora Cluster."
+  description = <<-EOT
+    If var.ssm_passwords_enabled is true, DB user passwords will be retrieved from SSM using `var.ssm_password_source` and the database username. If this value is not set, a default path will be created using the SSM path prefix and ID of the associated Aurora Cluster."
```
Nuru, line 16: Besides the formatting changes, it is good to break these long strings into multiple lines because they are not auto-wrapped when viewed on GitHub in code.
```suggestion
    If `var.ssm_passwords_enabled` is `true`, DB user passwords will be retrieved from SSM using
    `var.ssm_password_source` and the database username. If this value is not set,
    a default path will be created using the SSM path prefix and ID of the associated Aurora Cluster.
```

modules/aurora-mysql/variables.tf
```diff
@@ -161,25 +163,27 @@ variable "eks_component_names" {
 
 variable "replication_source_identifier" {
   type        = string
-  description = "ARN of a source DB cluster or DB instance if this DB cluster is to be created as a Read Replica. If this value is empty and replication is enabled, remote state will attempt to find a matching cluster in the Primary DB Cluster's region"
+  description = <<-EOT
+    ARN of a source DB cluster or DB instance if this DB cluster is to be created as a Read Replica. If this value is empty and replication is enabled, remote state will attempt to find a matching cluster in the Primary DB Cluster's region"
```
Nuru, line 167:
```suggestion
    ARN of a source DB cluster or DB instance if this DB cluster is to be created as a Read Replica. 
    If this value is empty and replication is enabled, remote state will attempt to find 
    a matching cluster in the Primary DB Cluster's region.
```

Nuru (changes requested):

modules/aurora-mysql-resources/modules/mysql-user/main.tf
milldr: removed db_user_key from SSM. It is redundant

modules/aurora-mysql-resources/modules/mysql-user/main.tf
milldr: this was originally commented out when starting the update, so I will uncomment and return this check

modules/aurora-mysql-resources/variables.tf
milldr: updated variable name and occurances

Nuru (approved):
Great work, @milldr

PR #478, opened 2022-09-01
Upstream `Datadog-Integration`

Benbentwo:
## what
* Upstreaming `datadog-integration` component
* **Added Variable** `context_host_and_filter_tags` instead of a hardcoded list in _main.tf_

## why
* Upstreaming code
* `context_host_and_filter_tags` lets reduce overhead configuration since we usually add stage, namespace and tenant to the host and filter tags of each account's integration


modules/datadog-integration/component.yaml
```diff
@@ -0,0 +1,38 @@
+# 'datadog-integration' component vendoring config
```
milldr, line 1: should we include `component.yaml` in upstreams?

milldr: we havent in the past, but it could be helpful

modules/datadog-integration/default.auto.tfvars
```diff
@@ -1,3 +1,6 @@
 # This file is included by default in terraform plans
 
-enabled = true
+enabled = false
+
+name = "datadog-integration"
```
milldr, line 5:
```suggestion
```

modules/datadog-integration/introspection.mixin.tf
```diff
@@ -0,0 +1,34 @@
+# <-- BEGIN DOC -->
```
milldr, line 1: remove introspection from upstream

modules/datadog-integration/component.yaml
Benbentwo: Yup, just removed it

PR #480, opened 2022-09-03
feat: adds initial cut of the ssm-parameters component

Gowiem:
## what
* Adds a new component: `ssm-parameters`

## why
* We use this component on Masterpoint projects to manage SSM Parameter Store resources in a GitOps-ish fashion.
* It enables managing secrets via sops and seeding them into SSM Params -- This be useful to reduce manual operations around secrets.
* It also supports managing normal SSM Params that are not secrets in the case that is useful for non-secret config. 

## references
* None

### DIFF
--- a/modules/ssm-parameters/README.md
+++ b/modules/ssm-parameters/README.md
@@ -31,14 +31,14 @@ components:
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.2.5 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 4.0 |
 | <a name="requirement_sops"></a> [sops](#requirement\_sops) | >= 0.5 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.0 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 4.0 |
 | <a name="provider_sops"></a> [sops](#provider\_sops) | >= 0.5 |
 
 ## Modules
--- a/modules/ssm-parameters/README.md
+++ b/modules/ssm-parameters/README.md
@@ -1,0 +1,100 @@
+# Component: `ss-parameters`
+
+This component is responsible for provisioning Parameter Store resources against AWS SSM. It supports normal parameter store resources that can be configured directly in YAML OR pulling secret values from a local Sops file.
+
+## Usage
+
+**Stack Level**: Regional
+
+Here are some example snippets for how to use this component:
+
+`stacks/dev/us-east-1.yaml` file:
+
+```yaml
+components:
+  terraform:
+    ssm-parameters:
+      vars:
+        sops_source_file: ../../config/secrets/dev.yaml
+        sops_source_key: ssm_params
+        params:
+          /DEV/TESTING:
+            value: This is a test of the emergency broadcast system.
+            description: This is a test.
+            overwrite: true
+            type: String
+```
+
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.2.5 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.0 |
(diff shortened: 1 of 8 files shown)


Gowiem:
@osterman @nitrocode -- adding this one onto the pile to see if it's of interest to you folks. Let me know your thoughts and if you'd like to see anything changed. 🤝

modules/ssm-parameters/README.md
```diff
@@ -0,0 +1,100 @@
+# Component: `ss-parameters`
```
nitrocode, line 1:
```suggestion
# Component: `ssm-parameters`
```

modules/ssm-parameters/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+This component is responsible for provisioning Parameter Store resources against AWS SSM. It supports normal parameter store resources that can be configured directly in YAML OR pulling secret values from a local Sops file.
```
nitrocode, line 3: Could this component be used for both SSM params and AWS Secrets Manager ? or would it be best to have a separate component for the latter ?

modules/ssm-parameters/versions.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_version = ">= 1.2.5"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 3.0"
```
nitrocode, line 7:
```suggestion
      version = ">= 4.0"
```

modules/ssm-parameters/main.tf
```diff
@@ -0,0 +1,35 @@
...
+  sops_yaml     = yamldecode(data.sops_file.source.raw)
+  secret_params = nonsensitive(local.sops_yaml[var.sops_source_key])
+
+  secret_params_normalized = {
+    for key, value in local.secret_params :
+    key => {
+      name        = key
+      value       = value
+      description = "SecureString param created from ssm-parameters component from SOPS source file: ${var.sops_source_file} from key at ${var.sops_source_key}"
+      overwrite   = true
```
nitrocode, line 11: Wouldn't it be safer to set this to `false` ?

nitrocode (dismissed):
Some nitpicks but this looks good to me

nitrocode:
Let's get one more reviewer tho. cc: @aknysh @osterman

modules/ssm-parameters/README.md
Gowiem: Ah I could definitely see it being upgraded for Secrets Manager usage as well. We typically don't use Secrets Manager as there aren't that many use-cases IMO, but I could definitely see the need for folks that use it! I like the idea... 

In that case, we could rename the component? Not sure how far you want to take this idea 😅

modules/ssm-parameters/main.tf
Gowiem: @nitrocode the point of this module is that your always managing your secrets via `SOPS` and then terraform is just injecting them into parameter store so systems that read from parameter store can have access to them. With that in mind, our thought process is that we always want the sops file to be the declarative source of truth and then this component should reconcile any out-of-band changes to the parameter store value that aren't made to the sops file. 

Does that make sense on why it's `true` here?

modules/ssm-parameters/versions.tf
```diff
@@ -0,0 +1,14 @@
...
+    }
+    sops = {
+      source  = "carlpett/sops"
+      version = ">= 0.5"
```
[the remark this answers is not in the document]
Gowiem: I see your point, but I have used this provider for years and haven't run into an issue. I would rather the flexibility to use `carlpett/sops` 0.6 and up (which are released). 

What about if we changed this to be `>= 0.5, < 1.0.`?

Gowiem, line 11:
```suggestion
      version = ">= 0.5, < 1.0"
```

modules/ssm-parameters/README.md
Gowiem, line 1:
```suggestion
# Component: `ssm-parameters`
```

modules/ssm-parameters/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "The ARN of a KMS key used to encrypt and decrypt SecretString values"
+}
+
+variable "params" {
+  type = map(object({
+    value       = string
+    description = string
+    overwrite   = bool
+    type        = string
+  }))
```
milldr, line 28: can we add a brief description for this variable?

milldr (approved):

modules/ssm-parameters/default.auto.tfvars
```diff
@@ -0,0 +1,3 @@
+# This file is included by default in terraform plans
```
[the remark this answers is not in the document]
milldr: I've removed that file

PR #488, opened 2022-09-09
`datadog-monitors` allow specific monitors to be disabled during merge

Benbentwo:
## what
* Allows `enabled: false` on monitors to succeed in merge

## why
* bugfix introduced from upstream a while ago

error:
```
│ Error: Inconsistent conditional result types
│ 
│   on main.tf line 92, in module "datadog_monitors_merge":
│   92:     (local.local_datadog_monitors_enabled) ? module.local_datadog_monitors_yaml_config.map_configs : {},
│     ├────────────────
│     │ local.local_datadog_monitors_enabled is true
│     │ module.local_datadog_monitors_yaml_config.map_configs is object with 38 attributes
│ 
│ The true and false result expressions must have consistent types. The 'true' value includes object attribute "aurora-replica-lag", which is absent in the 'false' value.
╵
```

### DIFF
--- a/modules/datadog-monitor/main.tf
+++ b/modules/datadog-monitor/main.tf
@@ -44,7 +44,7 @@ locals {
   datadog_monitors = {
     for k, v in module.datadog_monitors_merge :
     k => merge(v.merged, {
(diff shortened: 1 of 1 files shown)


mcalhoun (approved):
LGTM!

PR #489, opened 2022-09-09
Upstream argocd components

nitrocode:
## what
* Upstream argocd components

## why
* Upstream

## references
* N/A


modules/eks/argocd/main.tf
```diff
@@ -0,0 +1,241 @@
...
+  kubernetes_namespace = var.kubernetes_namespace
+  sops_directory       = abspath("./resources")
+  sops_file_name       = "argocd-notifications-secret.sops.yaml"
+}
+
+module "argocd" {
+  source  = "cloudposse/helm-release/aws"
+  version = "0.3.0"
+
+  name                   = "" # avoids hitting length restrictions on IAM Role names
```
Gowiem, line 114: @nitrocode what does your argocd chart install end up getting named? Due to the code [here](https://github.com/cloudposse/terraform-aws-helm-release/blob/master/main.tf#L41), we get funky names for our argocd installation like `argo-cd-argocd-server`, which I am not a fan of. What do your K8s resource names look like with this chart and have you found a way around that name duplication?

nitrocode: Ah that's a good question. Im unsure tbh

Gowiem: @nitrocode it's that you folks override `fullnameOverride` for all of your chart deployments I believe. This does away with whatever is passed in as the name 👍

osterman:
Superceeded by #560

PR #490, opened 2022-09-15
upstream `s3` component

Benbentwo:
## what
* Version Updates
* Adding variable to format logging bucket with our context

## why
* easier referencing of logging buckets with shared values


modules/s3-bucket/main.tf
```diff
@@ -9,6 +9,11 @@ locals {
   ]
 
   bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : data.template_file.bucket_policy.rendered
+
+  logging = var.logging != null ? {
+    bucket_name = var.logging_bucket_name_rendering_enabled ? format("%s-%s-%s-%s-%s", var.namespace, var.tenant, var.environment, var.stage, var.logging.bucket_name) : var.logging.bucket_name
+    prefix      = var.logging_bucket_name_rendering_enabled ? format("%s/%s/", var.logging.prefix, var.name) : var.logging.prefix
```
Nuru, line 15: As a general rule, we want to avoid hard-coded format strings (not that we have been widely successful at achieving that goal). Implement templates along the lines of account-map: https://github.com/cloudposse/terraform-aws-components/blob/dafde43d7ded86229d49492ebc85a754ad37e63d/modules/account-map/variables.tf#L41-L68

Nuru (changes requested):
Avoid hard coded format strings. Supply via template variables, use reasonable defaults.

Nuru (approved):

PR #493, opened 2022-09-15
s3-bucket: allow usage without providing a value for `tenant`

joe-niland:
## what
* Currently the s3-bucket module requires a value for `tenant`, either set via context or `account_map_tenant_name`. This change relaxes this requirement.

## why
* Some consumers of this module don't use tenant in resource naming

## references
* None

### DIFF
--- a/modules/s3-bucket/remote-state.tf
+++ b/modules/s3-bucket/remote-state.tf
@@ -5,7 +5,7 @@ module "account_map" {
   component   = "account-map"
   environment = var.account_map_environment_name
   stage       = var.account_map_stage_name
-  tenant      = coalesce(var.account_map_tenant_name, module.this.tenant)
+  tenant      = try(coalesce(var.account_map_tenant_name, module.this.tenant), "")
 
   context = module.this.context
 }
--- a/modules/s3-bucket/remote-state.tf
+++ b/modules/s3-bucket/remote-state.tf
@@ -5,7 +5,7 @@ module "account_map" {
   component   = "account-map"
   environment = var.account_map_environment_name
   stage       = var.account_map_stage_name
-  tenant      = try(coalesce(var.account_map_tenant_name, module.this.tenant), "")
+  tenant      = var.account_map_tenant_name
 
   context = module.this.context
 }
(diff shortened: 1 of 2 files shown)


Gowiem:
@Benbentwo @nitrocode -- Pulling you both in as you've got a lot of stake in this repo. @joe-niland ran into this issue as part of some Masterpoint work as we're not using `tenant` on a couple clients. We can confirm it works and I'd be happy to approve and merge, but figured I'd double check with you folks.

modules/s3-bucket/remote-state.tf
```diff
@@ -5,7 +5,7 @@ module "account_map" {
   component   = "account-map"
   environment = var.account_map_environment_name
   stage       = var.account_map_stage_name
-  tenant      = coalesce(var.account_map_tenant_name, module.this.tenant)
+  tenant      = try(coalesce(var.account_map_tenant_name, module.this.tenant), "")
```
nitrocode, line 8: Would this work instead since `module.this.tenant` is already provided by the `context` ?

```suggestion
  tenant      = var.account_map_tenant_name
```

Gowiem: @nitrocode I think it probably would. @joe-niland you mind testing this change out to confirm?

joe-niland: @nitrocode yes that does work, thanks.

I actually wasn't aware of [this behaviour](https://github.com/cloudposse/terraform-null-label/blob/master/main.tf#L51)!

I've modified the PR

nitrocode (approved):

nitrocode:
@joe-niland please fix the pre commit issues by locally running `make pr/auto-format`

joe-niland:
@nitrocode Not having much luck with that:
`make pr/auto-format` runs but doesn't change anything.
Running `pre-commit run --all-files` got the same error shown in the Actions output.

nitrocode:
Hmmm yes, maybe we do need to expand the exclusions list.

joe-niland:
@nitrocode I added a few more and it seemed to do the trick

nitrocode (approved):

nitrocode:
Thanks for the contribution @joe-niland !

PR #495, opened 2022-09-21
Allow IAM users to assume team roles via an optional flag

joe-niland:
## what
* Adds a flag `var.deny_all_iam_users` which, when disabled, will allow IAM users to assume team roles
* Specific users can still be denied via `var.denied_principal_arns`

## why
* The existing functionality denies all IAM users from assuming team roles
* In legacy/specific situations it would be good to allow this

## references
* None

(PR submitted on behalf of @Gowiem)


Gowiem:
@joe-niland 👍 You improved upon my original -- Thank you! 👏

Nuru (approved):

Gowiem:
Good stuff -- Thanks for the help @Nuru 👍

PR #497, opened 2022-09-26
upstream `eks` folder

Benbentwo:
## what
* Upstreams eks folder
* moves EKS components to eks folder.

## why
* latest updates and better formatting

## references
* closes #491 - superseded


Nuru (approved):
There is a bit of a regression in actions-runner-controller, and maybe elsewhere, but we will sort it out.

PR #501, opened 2022-10-06
VPC, DNS, CloudTrail, testing

Nuru:
## what

- Upstream changes to 
  - cloudtrail
  - cloudtrail-bucket
  - dns-delegated
  - dns-primary
  - vpc
- Run `bats` tests on modified modules
- Remove `default.auto.tfvars` per latest standard

## why

- VPC
  - Improve support for VPC Endpoints
  - Remove support for deprecated features
- DNS
  - Change default negative TTL to 60 seconds
  - Make SOA configurable in `dns-delegated`
  - Improve SOA documentation
- CloudTrail: update `remote-state`, fix provider version pinning
- Testing: enforce relevant standards from our Open Source Terraform modules

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -1,0 +1,54 @@
+name: bats
+
+on:
+  pull_request_target:
+    types: [labeled, opened, synchronize, unlabeled]
+
+jobs:
+  bats:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.state == 'open'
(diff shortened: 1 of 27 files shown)


Nuru:
The Bridgecrew errors are nuts. As far as I can tell, they are complaints about Ubutnu in general. There are separate complaints about 3 different versions of `go`, none of which are installed. I have added some suppressions to see if I can make these stop.

PR #502, opened 2022-10-06
GitHub OIDC, ECR

Nuru:
## what
- Create `github-oidc-iam-role` mixin with README
- Establish pattern for IAM Polices to go with `github-oidc-iam-role` and add policy for `ecr`
- Remove obsolete and now unused `root_account_tenant_name` from `ecr` and `vpc`
- Update `ecr` README to favor GitHub OIDC
- Fix reference to `module.introspection` in `github-oidc-provider`
- Update AWS version pins
- Removed unused files

## why
- GitHub OIDC is best practice, but we want to leave it as a mixin for now
- General cleanup


Nuru:
`bats` failure is due to syntax error in workflow in `master` branch, should be fixed by merging this PR.

PR #503, opened 2022-10-10
Spacelift updates

Nuru:
## what
- Update `spacelift` and `spacelift-worker-pool` components
- Update `account-map` and add `aws-config` to generate `aws-config-spacelift` file

## why

**Note:** There are is still a lot to be worked out, such as how to avoid the need to copy output from other modules into the `spacelift` configuration. This PR is not meant to stand as the final design. This is just moving the design forward incrementally.


Benbentwo (approved):
Looks good, @osterman's nitpick is valid, as it doesn't look backwards compatible, but since this is terraform-aws-components, we could always look through the history

Benbentwo (dismissed):

PR #507, opened 2022-10-12
Fix feature allowing IAM users to assume team roles

Nuru:
## what
- Replace `deny_all_iam_users` input with `iam_users_enabled`
- Fix implementation
- Provide more context for `bats` test failures

## why

- Cloud Posse style guide dictates that boolean feature flags have names ending with `_enabled`
- Previous implementation only removed 1 of 2 policy provisions that blocked IAM users from assuming a role, and therefore IAM users were still not allowed to assume a role. Since the previous implementation did not work, a breaking change (changing the variable name) does not need major warnings or a major version bump.
- Indication of what was being tested was too far removed from `bats` test failure message to be able to easily identify what module had failed

## notes

Currently, any component provisioned by SuperAdmin needs to have a special provider configuration that requires SuperAdmin to provision the component. This feature is part of what is needed to enable SuperAdmin (an IAM User) to work with "normal" provider configurations.  

## references

- Breaks change introduced in #495, but that didn't work anyway.


Gowiem (dismissed):
LGTM

PR #509, opened 2022-10-18
Update EKS basic components

Nuru:
## what && why

Update EKS cluster and basic Kubernetes components for better behavior on initial deployment and on `terraform destroy`.

- Update minimum Terraform version to 1.1.0 and use `one()` where applicable to manage resources that can be disabled with `count = 0` and for bug fixes regarding destroy behavior
- Update `terraform-aws-eks-cluster` to v2.5.0 for better destroy behavior
- Update all components'  (plus `account-map/modules/`)`remote-state` to v1.2.0 for better destroy behavior
- Update all components' `helm-release` to v0.7.0 and move namespace creation via Kubernetes provider into it to avoid race conditions regarding creating IAM roles, Namespaces, and deployments, and to delete namespaces when destroyed
- Update `alb-controller` to deploy a default IngressClass for central, obvious configuration of shared default ingress for services that do not have special needs. 
- Add `alb-controller-ingress-class` for the rare case when we want to deploy a non-default IngressClass outside of the component that will be using it
- Update `echo-server` to use the default IngressClass and not specify any configuration that affects other Ingresses, and remove dependence on `alb-controller-ingress-group` (which should be deprecated in favor of `alb-controller-ingress-class` and perhaps a specialized future `alb-controller-ingress`)
- Update `cert-manager` to remove `default.auto.tfvars` (which had a lot of settings) and add dependencies so that initial deployment succeeds in one `terraform apply` and destroy works in one `terraform destroy`
- Update `external-dns` to remove `default.auto.tfvars` (which had a lot of settings)
- Update `karpenter` to v0.18.0, fix/update IAM policy (README still needs work, but leaving that for another day)
- Update `karpenter-provisioner` to require Terraform 1.3 and make elements of the Provisioner configuration optional. Support block device mappings (previously broken). Avoid perpetual Terraform plan diff/drift caused by setting fields to `null`.
- Update `reloader`
- Update `mixins/provider-helm` to better support `terraform destroy` and to default the Kubernetes client authentication API version to `client.authentication.k8s.io/v1beta1`

## references

- https://github.com/cloudposse/terraform-aws-helm-release/pull/34
- https://github.com/cloudposse/terraform-aws-eks-cluster/pull/169
- https://github.com/cloudposse/terraform-yaml-stack-config/pull/56
- https://github.com/hashicorp/terraform/issues/32023


modules/eks/alb-controller-ingress-class/provider-helm.tf
```diff
@@ -0,0 +1,166 @@
+##################
```
milldr, line 1: can we add this file to `mixins/` with this PR?

modules/eks/cert-manager/main.tf
```diff
@@ -125,15 +118,15 @@ module "cert_manager_issuer" {
...
   chart_version        = var.cert_manager_issuer_chart_version
-  kubernetes_namespace = join("", kubernetes_namespace.default.*.id)
+  kubernetes_namespace = var.kubernetes_namespace
   create_namespace     = false
   wait                 = var.wait
   atomic               = var.atomic
   cleanup_on_fail      = var.cleanup_on_fail
   timeout              = var.timeout
 
   # Only install IAM role if letsencrypt_enabled is true
```
milldr, line 128: remove comment if no longer an option
```suggestion
```

modules/eks/cluster/github-actions-iam-policy.tf
```diff
@@ -55,19 +39,11 @@ data "aws_iam_policy_document" "github_actions_iam_policy" {
...
-  }
-  statement {
-    effect = "Allow"
-    actions = [
+      "ssm:GetParameters",
       "ssm:GetParametersByPath"
     ]
     resources = [
-      "*"
+      "arn:aws:ssm:*:*:parameter/platform/${module.eks_cluster.eks_cluster_id}/*"
```
milldr, line 46: Can we replace `platform` with `module.this.tenant` if available?

milldr (changes requested):
a couple of minor comments

modules/eks/cluster/github-actions-iam-policy.tf
Nuru: I want to get rid of this whole thing. It should be pulled in as a mixin/override, and I don't understand its use case. Meanwhile, I don't want to mess with it.

modules/eks/cert-manager/main.tf
Nuru: @milldr I do not understand your comment. The code comment is still accurate.

modules/eks/cluster/github-actions-iam-policy.tf
milldr: this was originally a mixin for a client, so it doesnt need to be upstream. We should remove this file and the related mixin

milldr (dismissed):

milldr (dismissed):

mixins/provider-helm.tf
```diff
@@ -114,14 +122,14 @@ provider "helm" {
...
-    token                  = local.kube_data_auth_enabled ? data.aws_eks_cluster_auth.eks[0].token : null
+    token                  = local.kube_data_auth_enabled ? one(data.aws_eks_cluster_auth.eks[*].token) : null
     # The Kubernetes provider will use information from KUBECONFIG if it exists, but if the default cluster
     # in KUBECONFIG is some other cluster, this will cause problems, so we override it always.
     config_path    = local.kubeconfig_file_enabled ? var.kubeconfig_file : ""
     config_context = var.kubeconfig_context
 
     dynamic "exec" {
-      for_each = local.kube_exec_auth_enabled ? ["exec"] : []
+      for_each = local.kube_exec_auth_enabled && length(local.certificate_authority_data) > 0 ? ["exec"] : []
```
Benbentwo, line 132: Couldn't this error out if `local.certificate_authority_data` is `null`? Since try only discards errors not null.

modules/eks/cert-manager/variables.tf
```diff
@@ -112,13 +124,14 @@ variable "cert_manager_issuer_values" {
 
 variable "create_namespace" {
   type        = bool
-  description = "Create the namespace if it does not yet exist. Defaults to `false`."
-  default     = null
+  description = "Create the namespace if it does not yet exist. Defaults to `true`."
+  default     = true
```
Benbentwo, line 128: 👍  since we now have a k8 resource tracking it in our module, can we apply this change to all variables.tf helm components in this PR (`alb-controller`, and the rest) and set it as the default way moving forward

modules/eks/echo-server/charts/echo-server/templates/ingress.yaml
```diff
@@ -9,20 +9,10 @@ metadata:
   name: {{ $fullName }}
   annotations:
     {{- if eq (printf "%v" .Values.ingress.nginx.enabled) "true" }}
-    kubernetes.io/ingress.class: {{ .Values.ingress.nginx.class }}
       {{- if (index .Values.ingress.nginx "tls_certificate_cluster_issuer") }}
     cert-manager.io/cluster-issuer: {{ .Values.ingress.nginx.tls_certificate_cluster_issuer }}
       {{- end }}
     {{- else if eq (printf "%v" .Values.ingress.alb.enabled) "true" }}
-    kubernetes.io/ingress.class: {{ .Values.ingress.alb.class }}
```
Benbentwo, line 17: why are we removing these ingress anotations?

modules/eks/echo-server/charts/echo-server/templates/ingress.yaml
```diff
@@ -9,20 +9,10 @@ metadata:
...
-      {{- if not .Values.ingress.alb.group_name }}
-    alb.ingress.kubernetes.io/load-balancer-name: {{ index .Values.ingress.alb "load_balancer_name" | default "k8s-common" }}
-      {{- end }}
-    alb.ingress.kubernetes.io/group.name: {{ index .Values.ingress.alb "group_name" | default "common" }}
```
Benbentwo, line 21: this allows us to test the alb-ingress-controller-group logs being forwarded to datadog

modules/eks/echo-server/main.tf
```diff
@@ -2,49 +2,32 @@ locals {
   enabled               = module.this.enabled
   ingress_nginx_enabled = var.ingress_type == "nginx" ? true : false
   ingress_alb_enabled   = var.ingress_type == "alb" ? true : false
-
-  alb_access_logs_enabled = var.alb_access_logs_enabled && var.alb_access_logs_s3_bucket_name != null && var.alb_access_logs_s3_bucket_name != ""
-  ingress_controller_group_enabled = var.enable_alb_controller_ingress_group ? [
```
Benbentwo, line 7: related to dd log forwarding test

mixins/provider-helm.tf
Nuru: Yes, but `local.certificate_authority_data` will never be `null` because module outputs are never `null`. If an output were to be set to `null`, it would be omitted rather than included with a value of `null`, so the `try()` will catch it.

modules/eks/cert-manager/variables.tf
Nuru: No. It is not uncommon to deploy external-dns and alb-controller to `kube-system` and we do not want to try to create that namespace. I only set it to true for `cert-manager` because it is a recommended security practice to keep `cert-manager` in an isolated namespace.

modules/eks/echo-server/charts/echo-server/templates/ingress.yaml
Nuru: We do not want `echo-server` to disturb or generate any configuration any other service might rely on. So we deploy it using defaults only (as much as we can). That way there will be no ripple effect when we remove it.

modules/eks/echo-server/charts/echo-server/templates/ingress.yaml
Nuru: But it also can cause the whole thing to be destroyed when `echo-server` is removed. We do not want that to be default behavior. Any such configuration should be done explicitly outside of `echo-server`, on the default Ingress.

modules/eks/echo-server/main.tf
Nuru: As above, Access Log settings are per Ingress group. We do not want `echo-server` being responsible for them. They should be set up via IngressClass in the `alb-controller` or other component.

PR #511, opened 2022-10-19
Update ECS cluster component to use new module

goruha:
## what
* Update ECS cluster component to use new module

## why
* Use capacity providers


modules/ecs/main.tf
```diff
@@ -27,47 +27,64 @@ module "target_group_label" {
...
-  #     s3_bucket_encryption_enabled   = true
-  #     s3_key_prefix                  = "/"
-  #   }
-  # }
-
-  setting {
-    name  = "containerInsights"
-    value = var.container_insights_enabled ? "enabled" : "disabled"
-  }
+resource "aws_security_group" "default" {
```
nitrocode, line 30: Let's use the latest sg module here

modules/ecs/README.md
```diff
@@ -25,6 +25,12 @@ components:
         # Create records will be created in each zone
         zone_names:
           - example.com
+        capacity_providers_fargate: true
+        capacity_providers_fargate_spot: true
+        capacity_providers_ec2:
+          default:
+            instance_type: t3.medium
+            max_size: 2
```
nitrocode, line 33: Please provide an ec2 example too for future use

modules/ecs/main.tf
```diff
@@ -3,7 +3,7 @@ locals {
```
nitrocode, line 3: IIRC before upstreaming changes to these components, we test them out first in a client.

Has this already been tested in a client before merging here?

PR #513, opened 2022-10-20
feat: adds the initial version of the imported-account component

Gowiem:
## what
* Adds the imported-account component	

## why
* Provides an easy means by which to create + manage the `OrganizationAccountAccessRole` for newly imported accounts. 

## references
* N/A

## Notes

* This is originally some work from @joe-niland that I've changed up a bit for a couple clients. Figuring it could be useful to others.
* I could see renaming this component to `organization-account-role` or similar. If there is anyone who agrees with me, feel free to speak up as I'm happy to do that.


modules/imported-account/README.md
```diff
@@ -0,0 +1,86 @@
...
+    imported-account:
+      # This component is to be run as OrgAdmin
+      backend:
+        s3:
+          role_arn: null
+      settings:
+        spacelift:
+          workspace_enabled: false
+      vars:
+        provision_org_access_role: true
```
joe-niland, line 29:
```suggestion
        profile: import-profile
        org_admin_arn: arn:aws:iam::123456789012:user/OrgAdmin
```

modules/imported-account/main.tf
```diff
@@ -0,0 +1,36 @@
+module "org_account_access_role" {
```
joe-niland, line 1: @Gowiem the original version had a role map var which allowed various roles to be created, with various policies. TBH how you have it is probably the most common use case, and the only way I've ever used it, but might be worth making this flexible, at least for the policies?

joe-niland (dismissed):
Thanks for adding this @Gowiem. Just a couple of minor comments.

modules/imported-account/main.tf
Gowiem: @joe-niland yeah, I figured keeping it to this single use-case better suited the `providers.tf` w/ `var.profile` usage because we should provision other, generic roles through a more generically named + standard `providers.tf` component than this one if we want to do that. I dunno though. Let's see what others think cause I'm fine with a different implementation thought process too.

modules/imported-account/README.md
```diff
@@ -0,0 +1,87 @@
+# Component: `imported-account`
+
+This component is optional and is used to create the `OrganizationAccountAccessRole` role in an imported account.
```
nitrocode, line 3: If the description is to simply add a role to a specific account, couldn't the name of this component be something like `iam-role` and allow role assumption override via `profile` / `import_role_arn` / etc and then we can make this component more generic?

modules/imported-account/main.tf
```diff
@@ -0,0 +1,36 @@
...
+  # This is a critical admin role, so we keep it locked down and enforce MFA
+  assume_role_conditions = [{
+    test     = "Bool"
+    variable = "aws:MultiFactorAuthPresent"
+    values   = ["true"]
+  }]
+
+  policy_document_count = 0
+  managed_policy_arns = [
+    "arn:aws:iam::aws:policy/AdministratorAccess"
```
nitrocode, line 32: data partition here

modules/imported-account/providers.tf
```diff
@@ -0,0 +1,4 @@
+provider "aws" {
+  profile = var.profile
```
nitrocode, line 2: We should use the standard `aws` provider in `providers.tf` so we can support both `profile` and explicit `role_arn` to assume, no?

modules/imported-account/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.0.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 2.0"
```
nitrocode, line 7:
```suggestion
      version = ">= 4.0"
```

Gowiem:
@goruha -- I'm coming back to this and reading @nitrocode 's [point here regarding "can't this just be an iam-role component?"](https://github.com/cloudposse/terraform-aws-components/pull/513#discussion_r1111056306) and I believe he is correct. We can accomplish the same work that we're doing with this purpose-built component by using the newer [`iam-role` component here](https://github.com/cloudposse/terraform-aws-components/tree/main/modules/iam-role). Closing this out as it is no needed 👍

PR #514, opened 2022-10-20
feat: introduces secrets mixin + usage in cognito component

Gowiem:
## what
* Adds the `secrets.mixin.tf` to the cognito component
* Updates cognito component to use secrets consumed from that mixin

## why
* This was an idea and experiment internally at Masterpoint that I wanted to float by you folks at Cloud Posse. The `secrets.mixin.tf` enables us to utilize interchangeable secret backends, which would help with component adoption and usage as it doesn't require such a prescriptive approach to passing secret values into components (which today largely requires pre-seeding SSM Parameter Store). 
* Updated cognito component as we needed to pass these values ("client_id" + "client_secret") for our usage. Since we use SOPS, we had to hack on this component and I thought this would be a good time to introduce this idea. 

## references
* N/A

## Notes
* Just looking for thoughts on this one -- It seems like a good pattern that could result in larger usage + adoption so I like it, but I could see it being too abstract for ya'll to support. 
* The one point I think that could be a good addition here was to require the incoming secrets. If we can do that, then we ensure that secrets that are required are provided and if not then we fail fast. I haven't looked into it fully, but it might be possible to do so via [`preconditions`](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions)? Just an idea for now. 

## Testing

* I have tested this with the following (both uncommented + commented): 

```yaml
        secret_mapping:
          - name: cognito_client_id
            path: cognito_client_id
            file: ../../config/secrets/global.yaml
            type: sops
          - name: cognito_client_secret
            path: cognito_client_secret
            file: ../../config/secrets/global.yaml
            type: sops
          # - name: cognito_client_id
          #   path: /cognito/client_id
          #   file: null
          #   type: ssm
          # - name: cognito_client_secret
          #   path: /cognito/client_secret
          #   file: null
          #   type: ssm
```


modules/cognito/versions.tf
```diff
@@ -8,5 +8,9 @@ terraform {
       # https://github.com/hashicorp/terraform-provider-aws/releases
       version = ">= 4.8.0"
     }
+    sops = {
+      source  = "carlpett/sops"
+      version = "~> 0.6"
```
joe-niland, line 13:
```suggestion
      version = ">= 0.7"
```

Gowiem:
@osterman @nitrocode @aknysh friendly ping on this -- would be great to get some feedback or gauge interest on this one when ya'll got the time!

modules/cognito/secrets.mixin.tf
```diff
@@ -0,0 +1,75 @@
+# secrets.mixin.tf provides an abstract way for components to consume secret values from external sources
```
nitrocode, line 1: We have decided to keep the mixins out of the component directories and instead put them in the root level [mixins](https://github.com/cloudposse/terraform-aws-components/tree/master/mixins) folder.

Gowiem: @nitrocode that makes sense, but in this case... the component code needs to be written with usage of *abstract* secrets in mind (i.e. something that will provide the `local.secrets` values).  Maybe each component that utilizes secrets could then have a default `secrets.tf` that provided the `local.secrets` values and that file could be left out or overwritten when `vendor pull` happens. 

Thoughts on that? I'd be happy to write up a "Component Development Guide" in this repo which includes this pattern so we'd have a path forward that we can consolidate on?

nitrocode: I think that can be built into the provider code. We really need a way to download a mixin and apply a patch or shell script through vendoring.

cc @aknysh @Nuru @osterman @mcalhoun

nitrocode (approved):
I'm OK with this change but it would be good to get a second opinion.

Thanks for the contribution @Gowiem . Please let us know if your `secrets.mixin.tf` has evolved since this PR was proposed.

cc: @aknysh @osterman

modules/cognito/secrets.mixin.tf
Gowiem: @nitrocode + @osterman -- Friendly ping to see if this PR is of interest. Totally fine if not, but just realized it lost steam and wanted to bring it back up to either move it forward or close it.

Gowiem:
Going to close this out as I don't foresee this going anywhere 👍

PR #523, opened 2022-11-10
feat: updates kms comp with embedded policy creation

Gowiem:
## what
* Updates the KMS module to support embedded policy creation

## why
* This allows for easy wiring in of `aws-team-role` roles into the KMS policy, so we do something like "Admins in the dev account have access to use this Key"

## references
* N/A


modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
+locals {
+  account_id        = data.aws_caller_identity.current.account_id
+  account_principal = "arn:aws:iam::${local.account_id}:root"
+
+  administration_principals = [
+    local.account_principal
+  ]
```
nitrocode, line 7: This local doesn't seem too useful

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
+locals {
+  account_id        = data.aws_caller_identity.current.account_id
+  account_principal = "arn:aws:iam::${local.account_id}:root"
```
nitrocode, line 3: Let's use the data source here for govcloud

```suggestion
  account_principal = "arn:${data.aws_partition.current.partition}:iam::${local.account_id}:root"
```

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
...
+
+    actions = [
+      "kms:*",
+    ]
+
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = local.administration_principals
```
nitrocode, line 57:
```suggestion
      identifiers = [local.account_principal]
```

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
+locals {
+  account_id        = data.aws_caller_identity.current.account_id
+  account_principal = "arn:aws:iam::${local.account_id}:root"
+
+  administration_principals = [
+    local.account_principal
+  ]
+
```
nitrocode, line 8: This local doesn't seem very useful

```suggestion
```

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
...
+  principals = sort(distinct(concat(
+    var.allowed_principal_arns,
+    module.allowed_role_map.principals,
+  )))
+}
+
+data "aws_caller_identity" "current" {}
```
nitrocode, line 15:
```suggestion
data "aws_caller_identity" "current" {}

data "aws_partition" "current" {}
```

modules/kms/variables.tf
```diff
@@ -3,6 +3,18 @@ variable "region" {
...
+variable "alias" {
+  type        = string
+  description = "The display name of the alias. The name must start with the word alias followed by a forward slash. If not specified, the alias name will be auto-generated."
+  default     = null
+}
+
+variable "description" {
+  type        = string
+  description = "The description for the KMS Key."
+  default     = null
```
nitrocode, line 15: If the description isn't set to the original value of `Parameter Store KMS master key`, won't that cause people's kms keys to be recreated?

modules/kms/variables.tf
```diff
@@ -15,16 +27,22 @@ variable "enable_key_rotation" {
   description = "Specifies whether key rotation is enabled"
 }
 
-variable "description" {
+variable "key_usage" {
   type        = string
-  default     = "Parameter Store KMS master key"
-  description = "The description of the key as viewed in AWS console"
+  default     = "ENCRYPT_DECRYPT"
+  description = "Specifies the intended use of the key. Valid values: `ENCRYPT_DECRYPT` or `SIGN_VERIFY`."
```
nitrocode, line 33: Should we add a validation rule here or is this already done by the upstream module?

modules/kms/variables.tf
```diff
@@ -15,16 +27,22 @@ variable "enable_key_rotation" {
...
+  type        = bool
+  default     = false
+  description = "Indicates whether the KMS key is a multi-Region (true) or regional (false) key."
+}
+
+variable "customer_master_key_spec" {
   type        = string
-  default     = ""
-  description = "The display name of the alias. The name must start with the word `alias` followed by a forward slash. If not specified, the alias name will be auto-generated."
+  default     = "SYMMETRIC_DEFAULT"
```
nitrocode, line 44: Is this a breaking change or is this the normal default and we're just being explicit?

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
...
+
+  statement {
+    sid    = "KeyAdministration"
+    effect = "Allow"
+
+    actions = [
+      "kms:*",
+    ]
+
+    resources = ["*"]
```
nitrocode, line 53: It would be good to allow overriding the `resources` here so users of this component can give granular permissions to `kms`. These would also allay warnings by bridgecrew shown above.

modules/kms/main.tf
```diff
@@ -1,15 +1,80 @@
...
+
+    principals {
+      type        = "AWS"
+      identifiers = local.principals
+    }
+  }
+}
```
nitrocode, line 80: Instead of hard coding this policy here, can we utilize this module and expose the inputs as variables so we can be as flexible as possible from the YAML?

https://github.com/cloudposse/terraform-aws-iam-policy

e.g.

```yaml
components:
  terraform:
    kms:
      vars:
        policy:
          iam_policy_statements:
            KeyAdministration:
              effect: "Allow"
              actions:
              - "kms:*"
              resources:
              - "*"
              principals:
                type: "AWS"
                # This may be something we can contribute to the upstream module to dynamically fill these %s in
                # or it can be added by the HCL for this component
                identifiers:
                - "arn:%s:iam::%s:root"
            KeyUsage:
              effect: "Allow"
              actions:
              - "kms:Encrypt"
              - "kms:Decrypt"
              - "kms:ReEncrypt*"
              - "kms:GenerateDataKey*"
              - "kms:DescribeKey"
              resources:
              - "*"
              principals:
                type: "AWS"
                # Same here, these can be provided by the HCL unless this `identifiers` key is provided
                identifiers:
                - "..."
```

modules/kms/variables.tf
Gowiem: I'm not sure honestly as it has been a long time, but updated regardless 😅

modules/kms/main.tf
Gowiem: While this is totally valid, it's non-trivial and others haven't done this work / needed this yet, so to move this PR along I'm going to consider this out of scope. If we decide we need this to merge, I can make the changes necessary 👍

Gowiem:
@nitrocode thanks for the review! Sorry it took me so long to circle back to it 😄

I couldn't complete this PR since I can't push to the branch I started here due to lack of permissions to push to `origin`. I've opened #1136 as a result. Closing this and we'll continue in the newer PR.

cc @goruha

PR #528, opened 2022-12-12
Upstream EKS Action Runner Controller

milldr:
## what
- Upstreaming the latest additions for the EKS actions runner controller component

## why
- We've added additional features for the ARC runners, primarily adding options for ephemeral storage and persistent storage. Persistent storage can be used to add image caching with EFS
- Allow for setting a `webhook_startup_timeout` value different than `scale_down_delay_seconds`. Defaults to `scale_down_delay_seconds`

## references
- N/A


modules/eks/actions-runner-controller/variables.tf
```diff
@@ -178,14 +178,16 @@ variable "runners" {
...
     labels                         = list(string)
-    storage                        = optional(string, "")
+    storage                        = optional(string, false)
+    pvc_enabled                    = optional(string, false)
     resources = object({
       limits = object({
         cpu               = string
         memory            = string
-        ephemeral_storage = optional(string, "")
+        ephemeral_storage = optional(string, false)
```
Benbentwo, line 190: shouldn't this be 
```suggestion
        ephemeral_storage = optional(string, "false")
```

?

milldr: It could be, but we would need to update the charts.

we pass these values directly to helm, which then determines if they should be used with something like
`{{- if .Values.pvc_enabled }}` or `{{- if and .Values.dind_enabled .Values.resources.limits.ephemeral_storage }}`
which looks for a boolean `true` or `false`. String would always be `true`

If we change this to `optional(string, "false")`, we would need to instead check for a string with helm instead

milldr: `optional(string, false)` works for this use case. Although is may be confusing that the default type is a `bool`, whereas the custom type requires a `string`

Benbentwo (approved):

PR #530, opened 2022-12-30
fix(aws-sso): add missing tf update perms

dudymas:
## what
* Changes for supporting [Refarch Scaffold](github.com/cloudposse/refarch-scaffold)
* TerraformUpdateAccess permission set added

## why
* Allow SSO users to update dynamodb/s3 for terraform backend

### DIFF
--- a/modules/aws-sso/variables.tf
+++ b/modules/aws-sso/variables.tf
@@ -3,6 +3,11 @@ variable "region" {
   description = "AWS Region"
 }
 
+variable "tfstate_environment_name" {
+  type        = string
+  description = "Moniker for environment with tfstate-backend deployed"
+}
+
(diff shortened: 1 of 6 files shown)


modules/aws-sso/variables.tf
```diff
@@ -3,6 +3,11 @@ variable "region" {
   description = "AWS Region"
 }
 
+variable "tfstate_environment_name" {
+  type        = string
+  description = "Moniker for environment with tfstate-backend deployed"
```
milldr, line 8:
```suggestion
  description = "The name of the stage where `tfstate-backend` is provisioned"
```

milldr (approved):
LGTM!

PR #548, opened 2023-02-03
`datadog-agent` allow values var merged

Benbentwo:
## what
* Allows values to be passed in and merged to values file

## why
* Need to be able to easily override values files


milldr (approved):

milldr (approved):

max-lobur (commented):
Thanks for catching this!

PR #551, opened 2023-02-06
team-assume-policy to look in auto for github-oidc-provider

Benbentwo:
## what
* team-assume-policy needs to look in auto stage for `github-oidc-provider`


mixins/github-actions-iam-role/github-actions-iam-role.mixin.tf
```diff
@@ -48,6 +48,8 @@ module "gha_assume_role" {
   trusted_github_repos = var.github_actions_allowed_repos
 
   context = module.gha_role_name.context
+
+  stage = "auto"
```
milldr, line 52: can this be a variable instead with default as "auto"? a few customers don't have the same stage names

PR #553, opened 2023-02-08
aws-sso: Add format to fill in role name in Identity roles

johncblandii:
## what

* Add `format` to the identity role description

## why

* The role has a `%s` without `format` wrapping it so the descriptions do not replace it with the role name


Benbentwo:
Closing as was solved by #567

PR #561, opened 2023-02-14
upstream `opsgenie-team`

Benbentwo:
## what
* Upstreams latest opsgenie-team component


modules/opsgenie-team/component.yaml
```diff
@@ -0,0 +1,39 @@
+# 'opsgenie-team' component vendoring config
```
milldr, line 1: remove `component.yaml` from upstream

modules/opsgenie-team/introspection.mixin.tf
```diff
@@ -1,6 +1,14 @@
+# <-- BEGIN DOC -->
```
milldr, line 1: remove `introspection` mixin from upstream

milldr (approved):

PR #566, opened 2023-02-21
alb-ingress-group tag filter update

Benbentwo:
## what
* update alb-ingress group tag filter

## why
* a group name is the stack tag value, which should be filtered for


milldr (approved):

Benbentwo:
Want to hold off on this, might be a red herring

PR #573, opened 2023-02-24
`spacelift-worker-pool` README update and AMI limits

johncblandii:
## what
* Limited the AMI to `x86_64`. This might not be required so can remove this from the PR, if necessary.
* Updated the README to include other required vars

## why
* The AMI search would retrieve arm64 and fail to deploy

### DIFF
--- a/modules/spacelift-worker-pool/README.md
+++ b/modules/spacelift-worker-pool/README.md
@@ -36,10 +36,15 @@ components:
       vars:
         enabled: true
         name: "spacelift-worker-pool"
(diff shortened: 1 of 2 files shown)


arcaven:
Please consider a variation of this in #576 because that filter approach may more easily be refactored into setting the architecture as an attribute in variables.tf in the near future, or setting it automatically based on the instance type indicated. Not trying to step on toes, I created the PR before I found this PR

johncblandii:
Fair enough. I'm good with that.

PR #579, opened 2023-02-28
datadog monitors improvements

Benbentwo:
## what
* Datadog monitor improvements
  * Prepends `(<stack slug>)` e.g. `(tenant-environment-stage)`
  * Fixes some messages that had improper syntax - dd uses `{{ var.name }}`

## why
* Datadog monitor improvements


dudymas (approved):
lgtm

PR #581, opened 2023-03-01
datadog-lambda-forwarder: if s3_buckets not set, module fails

kevcube:
This module attempts to do length() on the value for s3_buckets.

We are not using s3_buckets, and it defaults to null, so length() fails.


kevcube:
@Benbentwo tagging you as you seem to be active in this repository. Please also see #584

Benbentwo (changes requested):
This is a sync issue with the lambda forwarder which received an update https://github.com/cloudposse/terraform-aws-datadog-lambda-forwarder/pull/45. The variable in `1.1.0` which is the current module version is null (and checked null), whereas it is an empty array in `1.2.0`.

Can you bump the version to match 1.2.0 for the lambda module.

Benbentwo (approved):

Benbentwo:
LGTM thanks for the changes!

PR #583, opened 2023-03-02
bugfix: rds anomalies monitor not sending team information

Benbentwo:
## what
* Update monitor to have default CP tags


dudymas (approved):
lgtm

PR #584, opened 2023-03-02
remove unused code, appease tflint

kevcube:


modules/datadog-configuration/main.tf
```diff
@@ -1,11 +1,6 @@
 locals {
   enabled     = module.this.enabled
   asm_enabled = local.enabled && var.datadog_secrets_store_type == "ASM"
-  ssm_enabled = local.enabled && var.datadog_secrets_store_type == "SSM"
-
-  # https://docs.datadoghq.com/account_management/api-app-keys/
-  datadog_api_key = var.datadog_secrets_store_type == "ASM" ? data.aws_secretsmanager_secret_version.datadog_api_key[0].secret_string : data.aws_ssm_parameter.datadog_api_key[0].value
-  datadog_app_key = var.datadog_secrets_store_type == "ASM" ? data.aws_secretsmanager_secret_version.datadog_app_key[0].secret_string : data.aws_ssm_parameter.datadog_app_key[0].value
```
kevcube, line 8: these were never referenced

Benbentwo:
@kevcube is this PR a blocker for you? I'm hesitant to merge this in as eventually we want to support ASM, and some of the code here that's being removed will want to be added back in later.

If it is, I can look through and test this. However if it's for purely cosmetic purposes I'd like to hold off

kevcube:
@Benbentwo we have this component vendored in and it's setting off our linter - we can just ignore these files but I feel better about cleaning it up at the source rather than adding lint ignore rules

modules/datadog-configuration/modules/datadog_keys/variables.tf
```diff
@@ -1,16 +1,5 @@
-variable "region" {
-  type        = string
-  description = "AWS Region"
-}
-
```
Benbentwo, line 5: this needs to be reverted, is used by several components in the reference to it:
e.g.
```hcl
module "datadog_configuration" {
  source  = "../datadog-configuration/modules/datadog_keys"
  region  = var.region
  context = module.this.context
}
```

modules/datadog-configuration/modules/datadog_keys/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
```
Benbentwo, line 1: this file shouldn't be added as the provider lock provided by the calling component. 
e.g.
[datadog-monitors/versions.tf](https://github.com/cloudposse/terraform-aws-components/blob/master/modules/datadog-monitor/versions.tf)
and [it's calling file to this module](https://github.com/cloudposse/terraform-aws-components/blob/master/modules/datadog-monitor/provider-datadog.tf#L1-L5)

Benbentwo: we don't add versions here because then we'd have to keep them in sync rather than letting the calling module handle it

Benbentwo (changes requested):
few minor changes, looks like we can remove most of the vars except `region` and we'll want to remove the versions.tf.

Should significantly reduce the rules of the tflint on your end

modules/datadog-configuration/modules/datadog_keys/versions.tf
kevcube: but these modules can potentially be used independently, right? I understand not including a lockfile here or a `provider "aws" {}` but if there are resources in this submodule I think it is better to be explicit about the required version of provider, at least setting a minimum compatible version. it isn't required that it stays in sync with the parent module, as long as the version range is permissive.'

modules/datadog-configuration/modules/datadog_keys/variables.tf
kevcube: @Benbentwo I don't understand, `var.region` is not utilized anywhere in this `datadog_keys` module, so why does it need to be reverted?

Doesn't terraform only give a warning on passing an unused variable in to a child module? I think that's the correct way forward, then the warning is motivation to stop passing it because it is being unused.

PR #589, opened 2023-03-09
update `opsgenie-team` to be delete-able via `enabled: false`

Benbentwo:
## what
* Uses Datdaog Configuration as it's source of datadog variables
* Now supports `enabled: false` on a team to destroy it.


modules/opsgenie-team/provider-datadog.tf
```diff
@@ -0,0 +1,15 @@
+// This is a custom provider-datadog.tf because it is always enabled, this is because we always need the datadog provider to be configured, even if the module is disabled.
+
+module "datadog_configuration" {
+  source  = "../datadog-configuration/modules/datadog_keys"
+  region  = var.region
+  enabled = true
```
dudymas, line 6: surprising gotcha. I guess this means that the provider will error out until datadog is configured?

Benbentwo: datadog provider needs to be available even if the opsgenie team is disabled, reason being is we need the datadog provider to delete the opsgenie integration for that team

dudymas: makes sense!

max-lobur (approved):

PR #602, opened 2023-03-21
mq-broker: upgrade versions

johncblandii:
## what
* Upgraded to the latest `terraform-aws-mq-broker`
* Updated the `providers` to match the common pattern
* Updated the module versions

## why
* The component was dated


modules/mq-broker/main.tf
```diff
@@ -13,7 +13,7 @@ locals {
 
 module "mq_broker" {
   source  = "cloudposse/mq-broker/aws"
-  version = "0.14.0"
+  version = "2.0.1"
```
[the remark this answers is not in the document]
Nuru:
```suggestion
  version = "3.0.0"
```

johncblandii:
@aknysh all comments addressed.

milldr (approved):

modules/mq-broker/providers.tf
```diff
@@ -1,8 +1,14 @@
 provider "aws" {
```
milldr, line 1: reset this to match https://github.com/cloudposse/terraform-aws-components/blob/main/mixins/providers.depth-1.tf

milldr (changes requested):
resolve merge conflicts, but otherwise LGTM!

modules/mq-broker/providers.tf
goruha: Done

milldr (approved):

PR #603, opened 2023-03-21
Add Privileged Option for GH OIDC

milldr:
## what
- allow gh oidc role to use privileged as option for reading tf backend

## why
- If deploying GH OIDC with a component that needs to be applied with SuperAdmin (aws-teams) we need to set privileged here

## references
- https://cloudposse.slack.com/archives/C04N39YPVAS/p1679409325357119


mixins/github-actions-iam-role/github-actions-iam-role.mixin.tf
```diff
@@ -27,6 +27,11 @@ variable "github_actions_iam_role_attributes" {
   default     = []
 }
 
+variable "privileged" {
+  type        = bool
+  description = "True if the default provider already has access to the backend"
```
dudymas, line 32: I feel like the wording sounds less like a variable and more like an output. Maybe something more like this?
```suggestion
  description = "Used to indicate when the default provider already has access to the backend"
```

dudymas (dismissed):
Looks good, but I could nitpick about the variable. See what you think.

mixins/github-actions-iam-role/github-actions-iam-role.mixin.tf
Nuru: @dudymas Frankly, "used to indicate when" sounds just as much like an output to me, with the added detriment of being vague as to how it is used. At least with "True if", you know what to do with the input, even the description sounds like an output. 

To be honest, pretty much any concise description of a variable can be read as a description of an output, because it does, in fact, describe the output if the value of that variable is output. (The Terraform Providers documentation is full of examples.) Also keep in mind that space is limited in the documentation of the variable as it appears in the README, and that both the README and the Terraform code provide explicit indications this is an input not an output via context. However, if you have a stylistic suggestion to improve the descriptions in general without making them excessively verbose, please suggest it as an overall improvement to the Cloud Posse style guidelines.

In this particular case, the `privileged` input variable is one of Cloud Posse's standard variables that occurs in many modules, **so I think it is best to leave it with the same description we use elsewhere.**

Nuru (dismissed):

Nuru (approved):

PR #613, opened 2023-03-29
add providers to `mixins` folder

Benbentwo:
## what
*  Copies some common providers to the mixins folder

## why
* Have a central place where our common providers are held.

### DIFF
--- a/mixins/provider-datadog.tf
+++ b/mixins/provider-datadog.tf
@@ -1,0 +1,12 @@
+module "datadog_configuration" {
+  source  = "../datadog-configuration/modules/datadog_keys"
+  region  = var.region
+  context = module.this.context
+}
+
+provider "datadog" {
+  api_key  = module.datadog_configuration.datadog_api_key
+  app_key  = module.datadog_configuration.datadog_app_key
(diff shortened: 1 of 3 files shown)


max-lobur (approved):

mixins/providers-aws-superadmin.tf
```diff
@@ -0,0 +1,3 @@
+provider "aws" {
+  region = var.region
+}
```
osterman, line 3: Should `var.region` be defined?

mixins/provider-datadog.tf
```diff
@@ -0,0 +1,12 @@
+module "datadog_configuration" {
+  source  = "../datadog-configuration/modules/datadog_keys"
+  region  = var.region
```
osterman, line 3: Should `var.region` be defined?

osterman: If not, maybe add a comment, like:
```
# Depends on `var.region`
```

Benbentwo:
regarding `var.region` and it's necessity. I think this warrants a bigger discussion and isn't in scope of this PR which is a reference point for vendoring, of which our current providers use var.region

PR #614, opened 2023-03-29
Providers Mixin

milldr:
## what
Added `providers.tf` to mixins folder

## why
This file is the same across many components. When referring to it in documentation, we want to link a common location

## references
- https://github.com/cloudposse/refarch-scaffold/pull/107/


milldr:
duplicate of #613

PR #618, opened 2023-03-31
s3-bucket: use cloudposse template provider for arm64

### LINKED ISSUE #617: s3-bucket component does not include template provider in versions.tf
https://github.com/cloudposse/terraform-aws-components/blob/6da272530d01c449b1a886860ccaeac3573880ae/modules/s3-bucket/main.tf#L21-L27

This provider is used in code but not versioned.

I will make a PR to remove this template and instead use the "template" function because this component can't run on arm64 as it is.

kevcube:
## what
* use cloud posse's template provider

## why
* arm64
* also this provider was not pinned in versions.tf so that had to be fixed somehow

## references
* closes #617

### DIFF
--- a/modules/s3-bucket/main.tf
+++ b/modules/s3-bucket/main.tf
@@ -8,7 +8,7 @@ locals {
     format("arn:%s:iam::%s:root", local.aws_partition, module.account_map.outputs.full_account_map[acct])
   ]
 
-  bucket_policy = var.custom_policy_enabled ? data.aws_iam_policy_document.custom_policy[0].json : data.template_file.bucket_policy.rendered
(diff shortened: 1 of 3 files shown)


modules/s3-bucket/main.tf
```diff
@@ -18,14 +18,6 @@ locals {
 
 data "aws_partition" "current" {}
 
-data "template_file" "bucket_policy" {
```
osterman, line 21: No need to remove. We provide a arm binary. 

https://github.com/cloudposse/terraform-provider-template

https://github.com/cloudposse/terraform-aws-ecs-cloudwatch-sns-alarms/pull/34/files#diff-202419d570b64c1010bf146c14f4a9fa017a98b9c936d9af293c7802a8f79e98R9

kevcube: Gotcha will do

osterman (approved):

PR #622, opened 2023-04-04
docs: eks/actions-runner-controller webhook scaleUpTrigger config

dudymas:
## what
* update example config to use 30m for webhook_startup_timeout
* add docs on configuration for webhook scaleUpTriggers

## why
* runners will scale back down at the end of duration and could interrupt work
* while 30m is a recommended default, long durations may be necessary

## references
* [scaling runners](https://github.com/actions/actions-runner-controller/blob/master/docs/automatically-scaling-runners.md)

### DIFF
--- a/modules/eks/actions-runner-controller/README.md
+++ b/modules/eks/actions-runner-controller/README.md
@@ -238,11 +238,12 @@ the logs of the `actions-runner-controller-github-webhook-server` pod.
 
 ### Configuring Scaling
 
-The default scaleUpTrigger duration is 30 minutes. The time should be adjusted with `webhook_startup_timeout`
-to best fit your maximum job duration. If this time is too short, it can interrupt work since the autoscaler
-will assume any activity from the webhook has been resolved. Think of this as a way to tidy up jobs that might
-not finish cleanly or could be stuck. No particular time will fit all scenarios so you should expect to
-adjust this once you have a workload to test with.
+The default [`scaleUpTrigger`](https://github.com/actions/actions-runner-controller/blob/master/docs/automatically-scaling-runners.md#webhook-driven-scaling) duration is 30 minutes. The time should be adjusted with `webhook_startup_timeout`
+to best fit your maximum job duration. You should ensure the duration allows for
+backlog clearance. Project maintainers suggest "30m," but it depends on job lengths and maxReplicas.
+Use "30m" for low-backlog pools, and "2h30m" for high-backlog pools, like package building.
+Previously, "2m" was recommended, but it's insufficient. Update deployments to "30m" or consult with
+customers.
 
 You can read more about
 [scaling runners](https://github.com/actions/actions-runner-controller/blob/master/docs/automatically-scaling-runners.md)
--- a/modules/eks/actions-runner-controller/README.md
+++ b/modules/eks/actions-runner-controller/README.md
@@ -74,7 +74,7 @@ components:
                 cpu: 100m
                 memory: 128Mi
             webhook_driven_scaling_enabled: true
-            webhook_startup_timeout: "2m"
+            webhook_startup_timeout: "30m"
             pull_driven_scaling_enabled: false
             # Labels are not case-sensitive to GitHub, but *are* case-sensitive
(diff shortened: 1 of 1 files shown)


modules/eks/actions-runner-controller/README.md
```diff
@@ -236,6 +236,17 @@ After the webhook is created, select "edit" for the webhook and go to the "Recen
 (of a "ping" event) with a green check mark. If not, verify all the settings and consult
 the logs of the `actions-runner-controller-github-webhook-server` pod.
 
+### Configuring Scaling
+
+The default scaleUpTrigger duration is 30 minutes. The time should be adjusted with `webhook_startup_timeout`
```
osterman, line 241:
```suggestion
The default [`scaleUpTrigger`](https://github.com/actions/actions-runner-controller/blob/master/docs/automatically-scaling-runners.md#webhook-driven-scaling) duration is 30 minutes. The time should be adjusted with `webhook_startup_timeout`
```

osterman:
We're missing additional context provided by @Nuru . 

Please integrate the following. 

> This documentation update covers  It is crucial to set an appropriate value for the `webhook_startup_timeout` parameter when using the autoscaler with self-hosted GitHub Action runners,  to prevent job interruption and ensure smooth operation.
>
># Scale-Up Trigger and Webhook Startup Timeout
> The default scaleUpTrigger duration is 30 minutes. This time should be adjusted with webhook_startup_timeout to best fit > your maximum job duration. If this time is too short, it can interrupt work since the autoscaler will assume any activity from the webhook has been resolved. This parameter acts as a way to tidy up jobs that might not finish cleanly or could be stuck. No particular time will fit all scenarios, so you should expect to adjust this once you have a workload to test with.
>
># Proper Configuration for Webhook Startup Timeout
>The webhook_startup_timeout parameter is critical for autoscaling self-hosted GitHub Action runners. If the autoscaler cannot start the job within the specified amount of time, it will cancel its capacity reservation, causing the runner pool to scale down even though the job remains to be done. To avoid this issue, the webhook_startup_timeout should be set long enough that backlogged jobs in the queue can wait for other jobs in the queue to finish before being started.
>
>The project maintainers recommend a default value of "30m", but it really depends on the specific scenario, such as the job duration and likelihood of having more jobs queued than the maxReplicas. For pools that rarely expect backlogs, "30m" is sufficient. For pools with heavy workloads, such as building packages (which can have over 200 jobs launched practically at once), a value of "2h30m" is recommended.
>
>If your previously recommended setting was "2m", it is woefully inadequate. Please update existing deployments to "30m" or discuss with the customer to determine the most suitable value.
>
>Note for Runner Pools with MinReplicas = 0
>This configuration update is particularly critical for runner pools with minReplicas = 0, because in those pools, jobs could be left hanging indefinitely. In runner pools that maintain 1 or more idle runners, the idle runners will eventually get through all the jobs, but it will take a lot longer than if the runner pool were properly scaled.
>
># Further Reading
>For more information about scaling runners, please refer to the [official documentation](https://docs.example.com/scaling-runners).

Nuru (approved):

modules/eks/actions-runner-controller/README.md
```diff
@@ -236,6 +236,18 @@ After the webhook is created, select "edit" for the webhook and go to the "Recen
...
 the logs of the `actions-runner-controller-github-webhook-server` pod.
 
+### Configuring Scaling
+
+The default [`scaleUpTrigger`](https://github.com/actions/actions-runner-controller/blob/master/docs/automatically-scaling-runners.md#webhook-driven-scaling) duration is 30 minutes. The time should be adjusted with `webhook_startup_timeout`
+to best fit your maximum job duration. You should ensure the duration allows for
+backlog clearance. Project maintainers suggest "30m," but it depends on job lengths and maxReplicas.
+Use "30m" for low-backlog pools, and "2h30m" for high-backlog pools, like package building.
+Previously, "2m" was recommended, but it's insufficient. Update deployments to "30m" or consult with
+customers.
```
osterman, line 246: We shoudl not refer to "customers" here.

PR #626, opened 2023-04-05
update datadog_lambda_forwarder ref for darwin_arm64

kevcube:
## what
* update datadog-lambda-forwarder module for darwin_arm64

## why
* run on Darwin_arm64 hardware

### DIFF
--- a/modules/s3-bucket/versions.tf
+++ b/modules/s3-bucket/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 4.0"
+      version = ">= 4.0"
     }
     template = {
       source  = "cloudposse/template"
--- a/modules/s3-bucket/versions.tf
+++ b/modules/s3-bucket/versions.tf
@@ -8,7 +8,7 @@ terraform {
     }
     template = {
       source  = "cloudposse/template"
-      version = "2.2.0"
+      version = ">= 2.2.0"
     }
   }
 }
--- a/modules/alb/README.md
+++ b/modules/alb/README.md
@@ -23,7 +23,7 @@ components:
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0.0 |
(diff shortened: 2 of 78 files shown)


modules/s3-bucket/versions.tf
```diff
@@ -8,7 +8,7 @@ terraform {
     }
     template = {
       source  = "cloudposse/template"
-      version = "2.2.0"
+      version = ">= 2.2.0"
```
kevcube, line 11: idk, bats was failing over version pinning.

Gowiem:
@Nuru @milldr -- friendly ping here. We're seeing the `bats` workflow run bats tests for ALL of the components and not just those affected by this PR, which causes this PR to fail. [See this check for what I'm talking about](https://github.com/cloudposse/terraform-aws-components/actions/runs/4631351970/jobs/8194107122?pr=626). 

Ya'll are the two main contributors to https://github.com/cloudposse/terraform-aws-components/blob/master/.github/workflows/bats.yml, so wondering if you've seen that before and if that is a problem we need to deal with as part of this PR or if we should just ignore?

milldr:
@Gowiem Bats has a newer requirement for version pinning that is missing from some less recently updated components. Since you've updated `s3-bucket`, you'll need to resolve the checks for that component. Specifically this needs to be updated for `s3-bucket` `versions.tf`
```
    aws = {
      source  = "hashicorp/aws"
      version = ">= 4.0"
    }
```

milldr:
Pushing a PR now to resolve bats across all components. See #627

milldr:
#627 is merged now. Pls updated this branch

kevcube:
done @milldr

milldr:
It appears that since this PR is from a fork, `bats` is running for all components, so we have to have all resolved before we can merge this PR. Looks like we missed a few provider version pins on a few other components, so I'll push 1 more PR. Hold tight

Gowiem:
Thanks for the help @milldr 🥇

milldr:
1 more PR is needed 🙁

osterman (approved):

milldr:
we can merge this with the tests passing for now (although we'll have those resolved soon). thanks for the contribution!

kevcube:
thanks a bunch @milldr

PR #630, opened 2023-04-07
feat: s3-bucket, use templatefile for logging configs. closes #620

### LINKED ISSUE #620: Question
Have a question? Please checkout our [Slack Community](https://slack.cloudposse.com) or visit our [Slack Archive](https://archive.sweetops.com/). 

[Slack Community](https://slack.cloudposse.com)

## Describe the Feature

While using the [s3-bucket component](https://github.com/cloudposse/terraform-aws-components/tree/master/modules/s3-bucket) and trying to add logging to it, I want to use a `logging_bucket_name_rendering_template` format to work when null-label's `tenant` and `environment` are null.  When I tried using the suggested format for when only `tenant` is null, I ran into an error with terraform not being able to format a null value with `%s` -- this is consistent with [tf format syntax section](https://developer.hashicorp.com/terraform/language/functions/format#specification-syntax) docs (see last paragraph in that section).

See the example below.

I believe the required feature change is to remove any null values in [the bucket_name line](https://github.com/cloudposse/terraform-aws-components/blob/master/modules/s3-bucket/main.tf#LL14C63-L14C193), 

```sh
│ Error: Error in function call
│ 
│   on ../../modules/example/main.tf line 30, in locals:
│   30:     final_bucket_name = format("%s-%s-%s-%%s", var.test_namespace, var.test_tenant, var.test_environment, var.test_stage, var.test_logging_bucket_name)
│     ├────────────────
│     │ while calling format(format, args...)
│     │ var.test_environment is "myEnviornment"
│     │ var.test_logging_bucket_name is "myLoggingBucketName"
│     │ var.test_namespace is "myNameSpace"
│     │ var.test_stage is "myStage"
│     │ var.test_tenant is null
│ 
│ Call to function "format" failed: unsupported value for "%s" at 3: null value cannot be formatted.
╵
Releasing state lock. This may take a few moments...
```

## Expected Behavior
When null values are present in the null-label settings (eg, when `environment` is null), the format function should still generate a logging bucket name.

## Use Case
NA 

## Describe Ideal Solution
Adjust [this line](https://github.com/cloudposse/terraform-aws-components/blob/b5846c0e1d1312369cf11bafcfcb2a86688de03b/modules/s3-bucket/main.tf#LL14C63-L14C193) to remove nulls before calling the format function.

Something like this,
```
locals {
    non_null_values = compact([var.test_namespace, var.test_tenant, var.test_environment, var.test_stage, var.test_logging_bucket_name])
    final_bucket_name = format("%s-%s-%s-%%s", local.non_null_values)
}
```

If I can get feedback on a viable solution, I'm happy to submit a PR.

## Alternatives Considered
Manually set the logging bucket as needed.

## Additional Context

```
variable "test_namespace" {
  type        = string
  description = "The namespace to use for the test"
  default = "myNameSpace"
}
  
variable "test_tenant" {
  type        = string
  description = "The tenant to use for the test"
  default     = null
}

variable "test_environment" {
  type        = string
  description = "The environment to use for the test"
  default     = "myEnvironment"
}

variable "test_stage" {
  type        = string
  description = "The stage to use for the test"
  default     = "myStage"
}

variable "test_logging_bucket_name" {
    type        = string
    description = "The logging bucket name to use for the test"
}

locals {
    non_null_values = compact([var.test_namespace, var.test_tenant, var.test_environment, var.test_stage, var.test_logging_bucket_name])
    final_bucket_name = format("%s-%s-%s-%%s", local.non_null_values)
}

output "final_bucket_name" {
  value = local.final_bucket_name
}
```

westonplatter:
## what
* Changes scoped to `s3-bucket` component.
* For s3 bucket access-logging, use `templatefile` to generate the `logging.bucket_name` so that null values in [`var.namespace`, `var.tenant`, `var.environment`, `var.stage`, `var.logging.bucket_name`] can be ignored by passing in a different template file.
* The same `templatefile` approach is used for `logging.prefix` as well.

## why
* The previous approach with `format` was not able to handle `null` values.
* This is because `format` cannot ignore `null` values per its [docs](https://developer.hashicorp.com/terraform/language/functions/format#default-format-verbs), `Null values produce the string null if formatted with %v or %#v, and cause an error for other verbs.`. 
## references
* See example terraform highlighting issue with `format` on https://github.com/cloudposse/terraform-aws-components/issues/620
* `closes #620`


modules/s3-bucket/versions.tf
```diff
@@ -6,9 +6,5 @@ terraform {
       source  = "hashicorp/aws"
       version = ">= 4.0"
     }
-    template = {
-      source  = "cloudposse/template"
-      version = ">= 2.2.0"
-    }
```
kevcube, line 12: @westonplatter Sorry I was wrong about removing this, it is being used elsewhere 😣

westonplatter: all good. Seemed like a good idea to me as well. Reverted.

Nuru (changes requested):
@westonplatter Thank you for this contribution.

Although I am using the "Request changes" feature, I am not asking you to make changes, I am asking you to close/abandon this PR. 

The changes you make are breaking changes, causing everyone using non-default values to have to modify their code just to keep the existing infrastructure unmodified. We try very hard to avoid breaking changes because of the amount of disruption they cause and effort they require to adapt to.

Next, we want all customizations to be made available via Terraform inputs to the degree possible. This PR adds template files, which are difficult to override and difficult to manage. IMHO, they do not provide sufficient benefit to be worth the additional effort they require.

We already provide customizations for log bucket name rendering and log bucket prefix rendering, although I grant the descriptions for them are incorrect. If you want to open a new PR fixing those descriptions, that would be welcome.

For `logging_bucket_name_rendering_template`, Use `"%s-%[3]s-%[4]s-%[5]s"` when not using `tenant`. The value recommended in the current description is wrong. You could add that the arguments to the format string are, in order

```
var.namespace, var.tenant, var.environment, var.stage, var.logging.bucket_name
```

Similarly, the `logging_bucket_prefix_rendering_template` would be better described as a format function taking `var.logging.prefix, var.name` as arguments. 

Note also that all of this can be avoided by setting `var.logging_bucket_name_rendering_enabled` to `false` and setting `var.logging` to the desired bucket name and prefix.

So, since you can do what you need to do with the existing component, I would ask you not to make changes to it.

goruha:
@westonplatter as @Nuru provider workaround with `%s-%[3]s-%[4]s-%[5]s`  to solve initial problem.
Because of all the back-compatibility arguments, I'll close the PR.

PR #636, opened 2023-04-12
ecs-service: fix lint issues

kevcube:


modules/ecs-service/remote-state.tf
```diff
@@ -72,6 +72,7 @@ module "ecs_cluster" {
 }
 
 # This is purely a check to ensure this zone exists
+# tflint-ignore: terraform_unused_declarations
```
kevcube, line 75: I'd like if this data source were just removed but idk what it's used for.

milldr (approved):

PR #638, opened 2023-04-13
Athena CloudTrail Queries

milldr:
## what
- added cloudtrail integration to athena
- conditionally allow audit account to decrypt kms key used for cloudtrail

## why
- allow queries against cloudtrail logs from a centralized account (audit)

## references
n/a


modules/athena/main.tf
```diff
@@ -13,7 +13,6 @@ module "athena" {
   enforce_workgroup_configuration    = var.enforce_workgroup_configuration
   publish_cloudwatch_metrics_enabled = var.publish_cloudwatch_metrics_enabled
   workgroup_encryption_option        = var.workgroup_encryption_option
-  encryption_option                  = var.encryption_option
```
milldr, line 16: this is deprecated, dne in module

modules/cloudtrail/cloudtrail-kms-key.tf
```diff
@@ -1,3 +1,8 @@
+locals {
+  audit_access_enabled = module.this.enabled && var.audit_access_enabled
+  audit_account_id     = module.account_map.outputs.full_account_map[var.audit_account_name]
```
Nuru, line 3:
```suggestion
  audit_account_id     = module.account_map.outputs.full_account_map[module.account_map.outputs.audit_account_account_name]
```

Nuru: `module.account_map` is missing

modules/cloudtrail/variables.tf
```diff
@@ -59,3 +65,15 @@ variable "is_organization_trail" {
...
+  type        = bool
+  default     = false
+  description = "If `true`, allows the Audit account access to read Cloudtrail logs directly from S3. This is a requirement for running Athena queries in the Audit account."
+}
+
+variable "audit_account_name" {
+  type        = string
+  default     = "core-audit"
+  description = "The key used in Account Map to find the Audit account"
+}
```
Nuru, line 79: Not needed:

```suggestion
```

PR #643, opened 2023-04-18
feat: cloudtrail-bucket can have acl configured

dudymas:
## what
* add `acl` var to `cloudtrail-bucket` component

## why
* Creating new cloudtrail buckets will fail if the acl isn't set to private

## references
* This is part of [a security update from AWS](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-faq.html)


milldr (approved):

Nuru:
@dudymas @milldr 

ACLs are depreciated in favor of "S3 Object Ownership". We should not be setting the `log-delivery-write` ACL, we should be updating the bucket policy. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html#grant-log-delivery-permissions-general

PR #652, opened 2023-04-28
Add DNS Delegated Component Name Variable to Aurora

milldr:
## what
set dns-delegated as variable in remote state

## why
support customer with different dns-delegated component name

## references
n/a


Nuru (changes requested):
Get DNS from `account-map` or else configure DNS zone name directly

PR #663, opened 2023-05-10
Add `iam-policy` to `ecs-service`

milldr:
## what
Add an option to attach the `iam-policy` resource to `ecs-service`

## why
This policy is already created, but is missing its attachment. We should attach this to the resource when enabled

## references
https://cloudposse.slack.com/archives/CA4TC65HS/p1683729972134479


goruha (approved):

goruha (changes requested):
fix bats

goruha (approved):

PR #666, opened 2023-05-10
upstream `acm` and `datadog-integration`

Benbentwo:
## what
* ACM allows disabling `*.my.domain`
* Datadog-Integration supports allow-list'ing regions


modules/acm/main.tf
```diff
@@ -22,7 +22,7 @@ module "acm" {
   domain_name                       = var.domain_name
   process_domain_validation_options = var.process_domain_validation_options
   ttl                               = 300
-  subject_alternative_names         = concat([format("*.%s", var.domain_name)], var.subject_alternative_names)
+  subject_alternative_names         = concat(var.enable_asterisk_subject_alternative_name ? [format("*.%s", var.domain_name)] : [], var.subject_alternative_names)
```
milldr, line 25:
```suggestion
  subject_alternative_names         = var.enable_asterisk_subject_alternative_name ? concat([format("*.%s", var.domain_name)] : [], var.subject_alternative_names)
```

Benbentwo: that's not the point of this variable. that would mean if `enable_asterisk_subject_alternative_name` is false, there are no _subject_alternative_names_, we instead want to only remove `*.foo.bar` and still allow other subject_alternative_names to be added

milldr (approved):

PR #685, opened 2023-05-19
Transit Gateway `var.connections` Redesign

milldr:
## what
- Updated how the connection variables for `tgw/hub` and `tgw/spoke` are defined
- Moved the old versions of `tgw` to `deprecated/tgw`

## why
- We want to be able to define multiple or alternately named `vpc` or `eks/cluster` components for both hub and spoke
- The cross-region components are not updated yet with this new design, since the current customers requesting these updates do not need cross-region access at this time. But we want to still support the old design s.t. customers using cross-region components can access the old components. We will need to update the cross-region components with follow up effort 

## references
- https://github.com/cloudposse/knowledge-base/discussions/112


Benbentwo:
Has this been tested on a working example?

milldr:
> Has this been tested on a working example?

Yes in SM

Benbentwo (approved):
This is a breaking change for cross-region spokes, however, overall this looks like an improvement. Approving with the internal discussion that we will create a ticket to refactor and re-support cross-region spokes, in the meantime, clients can use the `deprecated/tgw` component for cross region functionality

PR #693, opened 2023-05-24
feat: add group management to aws-sso

dudymas:
## what
* Specify Identity Center groups to manage with terraform

## why
* A better alternative to click-ops'ing the groups into AWS Identity Center


dudymas:
closing for now in favor of ssosync taking over this

PR #695, opened 2023-05-24
Made AWS Config respect privileged argument

zdmytriv:
## what
* Made AWS Config respect privileged argument

## why
* While applying to some accounts like `root` or `identity` user should be `SuperAdmin`. Once user is SuperAdmin remote-state will not be pulled. In this case variable `var.privileged=true` fixes this issue.


Benbentwo (approved):
LGTM

PR #699, opened 2023-05-25
EKS FAQ for Addons

milldr:
## what
Added docs for EKS Cluster Addons

## why
FAQ, requested for documentation

## references
DEV-846


modules/eks/cluster/variables.tf
```diff
@@ -398,3 +398,12 @@ variable "addons" {
   description = "Manages [`aws_eks_addon`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon) resources"
   default     = []
 }
+
+variable "addons_depends_on" {
+  type        = bool
+  description = <<-EOT
+    If provided, all addons will depend on node groups and therefore not be installed until nodes are finalized.
```
Nuru, line 405:
```suggestion
    If set `true`, all addons will depend on managed node groups provisioned by this component and therefore not be installed until nodes are provisioned.
```

Nuru (changes requested):
Not sure exactly what to say, but we should probably make some mention that in our current standard deployment with no node groups and just Karpenter doing the provisioning, Add Ons like Core DNS that require a node to be provisioned first will simply not work, despite `addons_depends_on`.

modules/eks/cluster/variables.tf
milldr: Added a warning that addons may not work with Karpenter to the section in the readme

Nuru (approved):

PR #706, opened 2023-06-02
Combined `guardduty/common` and `guardduty/root` components into single `guardduty` component

zdmytriv:
## what
* Combined `guardduty/common` and `guardduty/root` components into single `guardduty` component

## why
* It supposes to make it easier to maintain and use as single component


milldr:
done by https://github.com/cloudposse/terraform-aws-components/pull/725

PR #709, opened 2023-06-04
Combined `securityhub/common` and `securityhub/root` components into single `securityhub` component

zdmytriv:
## what
* Combined securityhub/common and securityhub/root components into single securityhub component

## why
* It supposes to make it easier to maintain and use as single component


modules/securityhub/README.md
```diff
@@ -33,62 +33,55 @@ The example snippet below shows how to use this component:
...
 
-1. Apply `securityhub/common` to all accounts
-2. Apply `securityhub/root` to `core-root` account
-3. Apply `securityhub/common` to `core-security` with `var.admin_delegated = true`
+In order to deploy AWS Security Hub to multiple accounts with single aggregation region.
+
+1. Deploy `securityhub` to every account except `core-security` and `core-root`. This will deploy Security Hub account resource.
+2. Deploy `securityhub` to `core-security` with `var.admin_delegated = false`. This will deploy Security Hub account resource into `core-security`
+3. Deploy `securityhub` to `core-root` (use `SuperAdmin` role). . This will deploy Security Hub into `core-root`.
+4. Deploy `securityhub` to `core-security` with `var.admin_delegated = true`. This will delegate `core-security` as a collector account in particular region so all findings will be reported into it.
```
Nuru, line 59: This is not reasonable. Order of operations should be:

1. Deploy to `core-root` using a role authorized to make changes in the root account. (Which roles are authorized will vary based from org to org based on how much the root account has been locked down.) This will deploy Security Hub account resource into `core-root` and delegate Security Hub administration to `core-security`
2. Deploy to `core-security`. This will designate `core-security` as the collector account for all regions so all findings will be reported into it. It will also enable Security Hub in all current accounts and set up new accounts to be automatically enabled.

If the above is not how it works, it should be. My understanding is that was the point of [awsutils_security_hub_organization_settings](https://registry.terraform.io/providers/cloudposse/awsutils/latest/docs/resources/security_hub_organization_settings#member_accounts).

modules/securityhub/main.tf
```diff
@@ -58,3 +62,9 @@ resource "awsutils_security_hub_organization_settings" "this" {
   member_accounts          = local.member_account_list
   auto_enable_new_accounts = true
 }
+
+resource "aws_securityhub_organization_admin_account" "this" {
+  count = local.enabled && local.organization_admin_account ? 1 : 0
```
Nuru, line 67:
```suggestion
  count = local.enabled && local.is_organization_admin_account ? 1 : 0
```

modules/securityhub/providers.tf
```diff
@@ -3,7 +3,7 @@ provider "aws" {
 
   profile = module.iam_roles.profiles_enabled ? coalesce(var.import_profile_name, module.iam_roles.terraform_profile_name) : null
   dynamic "assume_role" {
-    for_each = module.iam_roles.profiles_enabled ? [] : ["role"]
+    for_each = var.privileged || module.iam_roles.profiles_enabled ? [] : ["role"]
```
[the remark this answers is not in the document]
Nuru: This is not how `var.privileged` is intended to be used. Its meaning is that the current user has access to the Terraform state backend, so `remote-state` should not assume a role when fetching remote state. It should be passed into `remote-state` and `iam-roles`. 

`var.privileged` will be removed soon with an update to `account-map`, but until then, I suppose it is OK to use as an escape hatch for SuperAdmin. 
- Make sure to start with the updated `providers.tf` from https://github.com/cloudposse/terraform-aws-components/pull/702
- Use `var.privileged` to override both the role and the profile
- Pass `var.privileged` into `iam-roles`

We can remove `var.privileged` later, as part of a larger update.

Nuru (changes requested):

modules/securityhub/README.md
zdmytriv: In 1.
> This will deploy Security Hub account resource into core-root and delegate Security Hub administration to core-security
This will deploy Security Hub into root account and as well will enable Security Hub in delegated (in our example `core-security`) account.

So it is better to deploy to delegated account (`core-security`) first and then deploy to root account.

I'm designating core-security as main collector account in step 4 when all other accounts are set up.

modules/securityhub/README.md
```diff
@@ -33,62 +33,51 @@ The example snippet below shows how to use this component:
...
 ## Deployment
 
-1. Apply `securityhub/common` to all accounts
-2. Apply `securityhub/root` to `core-root` account
-3. Apply `securityhub/common` to `core-security` with `var.admin_delegated = true`
+In order to deploy AWS Security Hub to multiple accounts with single collection region.
+
+1. Deploy `securityhub` to every AWS account/region except organization root account.
+2. Deploy `securityhub` to organization root account (use `SuperAdmin` role).
+4. Update `securityhub` by designating one account as main collector account.
```
Nuru, line 58: Why does this need to be 3 steps? Shouldn't we designate the collector (in stacks YAML) and then deploy to root to do the delegation and then deploy everywhere else?

modules/securityhub/variables.tf
```diff
@@ -37,6 +37,11 @@ variable "central_resource_collector_region" {
   type        = string
 }
 
+variable "organization_admin_account" {
+  description = "The name of the organization admin account"
+  type        = string
+}
+
```
Nuru, line 44: This should not be required. We could take it out, or leave it in so it could be overridden, but it should at least default to `account_map.outputs.root_account_account_name`.

Nuru (changes requested):

modules/securityhub/variables.tf
zdmytriv: fixed

milldr:
done by https://github.com/cloudposse/terraform-aws-components/pull/728

PR #710, opened 2023-06-05
Feat/argocd repo/add template repos

dudymas:
## what
* allow argocd-repo component to skip creating the repo
* allow argocd-repo component to manage secrets in a repo

## why
* It's difficult to set up argocd repos correctly
* make it easier to rotate a secret such as a PAT on a set of applications


dudymas:
Will work on separate component rather than modifying argocd-repo

dudymas:
re-opening for discussion/refresh

dudymas:
re-closing since this should still be a separate module at this point

PR #719, opened 2023-06-09
fix: correct aws-ssosync outputs and logging

dudymas:
## what
* aws-ssosync
  * add log-group
  * add service policy
  * adjust outputs to support enabled flag

## why
* won't deploy without output changes
* logging needs a group and permissions to add to it

## references
* [Notes on the
service-role](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AWSLambdaBasicExecutionRole.html)
  * Yes, the resources is a '*', so I'd be okay with potentially changing this
  to be finer grained if folks feel like logging should be more secure around
  user syncs

### DIFF
--- a/modules/aws-ssosync/.gitignore
+++ b/modules/aws-ssosync/.gitignore
@@ -1,0 +1,3 @@
+*.zip
+*.tar.gz
+dist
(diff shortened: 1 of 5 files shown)


dudymas:
closing as stale

PR #720, opened 2023-06-09
Add `dns-primary` to `eks/cert-manager`

milldr:
## what
- Added dns-primary zones to cert-manager

## why
- We want to allow cert-manager to update Route 53 Hosted Zone records in the primary zones as well. This permission is required for Cert Manager to issue certificates for the primary domains. For example, when `var.letsencrypt_enabled` to set to `true`

## references
- [internal slack reference](https://cloudposse.slack.com/archives/C04N39YPVAS/p1686267088255179?thread_ts=1686265911.001219&cid=C04N39YPVAS)


Nuru:
@milldr I am closing this because I don't think it works, is necessary, or is a good idea. Of course, I could be wrong about all of that, in which case you can reopen this, but let's talk first.

Getting `cert-manager` to work cross-account will require setting up roles in the target account that `cert-manager` can assume, then configuring the solvers to use that role. This PR does not implement any of that.

Furthermore, it is not a good idea. The reason for having DNS zones in different accounts is to limit access. Providing this kind of cross-account access defeats that. 

"primary" zones can be deployed to any account. If an account needs access to an apex domain, that domain should be deployed as "primary" in that account. 

We can talk more about this if you want.

PR #724, opened 2023-06-15
Added s3 mirroring support

goruha:
## what
* Upstream ECS task definition mirroring to S3 bucket


goruha:
Close the PR as it is not actual anymore.
We have s3 mirroring merged in #810

PR #730, opened 2023-06-21
[account-map] Feature flag to enable legacy Terraform role mapping

Nuru:
## what

- [account-map] Add `legacy_terraform_uses_admin` feature flag to retain backwards compatibility

## why

- Historically, the `terraform` roles in `root` and `identity` were not used for Terraform plan/apply, but for other things, and so the `terraform_roles` map output selected the `admin` roles for those accounts. This "wart" has been remove in current `aws-team-roles` and `tfstate-backend` configurations, but for people who do not want to migrate to the new conventions, this feature flag enables them to maintain the status quo with respect to role usage while taking advantage of other updates to `account-map` and other components.

## references

This update is recommended for all customers wanting to use ***any*** component version 1.227 or later. 

- #715
-


mcalhoun (approved):
LGTM!

PR #739, opened 2023-06-28
Added policy for extending Poweruser for `aws-teams`

milldr:
## what
- Added a policy to extend the Poweruser access for the `devops` team in `core-identity`

## why
- Since we are no longer using `SuperAdmin` to apply `github-oidc-provider` across the accounts, we need permission to apply the component in each account using the given AWS IAM role. In all accounts except `core-identity`, this is already Admin, so this is a non problem. 
- However in `core-identity` only, we dont want to do that because we dont want to give `devops` the ability to lock itself out in the middle of an apply, and PowerUser alone is not enough.
- Add only the permission we need to the `devops` team. I created an `extended_poweruser` policy that adds only what we need to plan and apply, and I attached that to the `devops` team in `core-identity`

## references
- n/a

### DIFF
--- a/modules/aws-teams/policy-extended-poweruser.tf
+++ b/modules/aws-teams/policy-extended-poweruser.tf
@@ -1,4 +1,4 @@
-# This policy adds additional permission required to apply a limit set of Terraform
+# This policy adds additional permission required to apply a limited set of Terraform
 # resources in the `core-identity` account. We do not want to grant the full
 # admin access to `core-identity` so that we do not allow users to lock themselves out.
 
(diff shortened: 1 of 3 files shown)


modules/aws-teams/policy-extended-poweruser.tf
```diff
@@ -0,0 +1,33 @@
+# This policy adds additional permission required to apply a limit set of Terraform
```
milldr, line 1:
```suggestion
# This policy adds additional permission required to apply a limited set of Terraform
```

Nuru:
I am closing this as I am not a fan of expanding the poweruser role in identity. It is really specific to the github-oidc-provider update, which is tech debt we will pay for over and over. We can reopen it if we do not agree on a better solution.


Customers need to make a choice about who can modify root and identity. I wanted it to be limited to SuperAdmin, but lost that fight badly, so the current aws-teams in refarch has a managers team that has Admin access in both root and identity. Combined with dynamic Terraform roles, this allows both SuperAdmin and identity-managers to make changes in Identity.

PR #773, opened 2023-07-23
Update `alb` component

aknysh:
## what
* Update `alb` component

## why
* Fixes after provisioning and testing on AWS


modules/alb/variables.tf
```diff
@@ -222,6 +222,12 @@ variable "dns_delegated_component_name" {
   description = "Atmos `dns-delegated` component name"
 }
 
+variable "dns_delegated_environment_name" {
+  type        = string
+  default     = "gbl"
```
Nuru, line 227: I would like this to default to `null`, and when null, use `module.iam_roles.global_environment_name`.

Nuru (changes requested):

Nuru (approved):

PR #774, opened 2023-07-24
Update Variable Types for Cloudtrail Buckets

RoseSecurity:


RoseSecurity (commented):
Thank you for making these updates!

PR #783, opened 2023-07-28
`bastion` support for `availability_zones` and public IP and subnets

milldr:
## what
- Add support for `availability_zones`
- Fix issue with public IP and subnets
- `tflint` requirements -- removed all unused locals, variables, formatting

## why
- All instance types are not available in all AZs in a region
- Bug fix

## references
- [Internal Slack reference](https://cloudposse.slack.com/archives/C048LCN8LKT/p1689085395494969)


dudymas (approved):

dudymas:
In the future it would be best to use a `instance_type_offerings` query: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ec2_instance_type_offerings

This works just fine, though.

PR #790, opened 2023-08-03
Feat/add macie

mcalhoun:
## what
* Add the AWS Macie component

## why
* To upstream the component developed for a customer


modules/macie/remote-state.tf
```diff
@@ -0,0 +1,12 @@
+module "account_map" {
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "1.4.3"
```
milldr, line 3:
```suggestion
  version = "1.5.0"
```

modules/macie/providers.tf
```diff
@@ -0,0 +1,42 @@
...
+  type        = string
+  default     = null
+  description = "AWS Profile name to use when importing a resource"
+}
+
+variable "import_role_arn" {
+  type        = string
+  default     = null
+  description = "IAM Role ARN to use when importing a resource"
+}
```
milldr, line 42:
```suggestion
provider "aws" {
  region = var.region

  # Profile is deprecated in favor of terraform_role_arn. When profiles are not in use, terraform_profile_name is null.
  profile = module.iam_roles.terraform_profile_name

  dynamic "assume_role" {
    # module.iam_roles.terraform_role_arn may be null, in which case do not assume a role.
    for_each = compact([module.iam_roles.terraform_role_arn])
    content {
      role_arn = assume_role.value
    }
  }
}

module "iam_roles" {
  source  = "../account-map/modules/iam-roles"
  context = module.this.context
}
```

updated providers recommendation. See [providers mixin](https://github.com/cloudposse/terraform-aws-components/blob/main/mixins/providers.depth-1.tf)

milldr (commented):
updated component requirements

PR #791, opened 2023-08-03
add additional waf features

mcalhoun:
## what
* Add the ability to specify a list of ALBs to attach WAF to
* Add the ability to specify a list of tags to target ALBs to attach WAF to

## why
* To provider greater flexibility in attaching WAF to ALBs


modules/waf/remote-state.tf
milldr, line 1: delete empty file

milldr (commented):
remove empty file, otherwise LGTM

modules/waf/remote-state.tf
goruha: This is not an empty file. This is just a file with only permissions changed.

milldr (approved):

Issue #797, opened 2023-08-07
first deployment of eks/cluster will never work


sgtoj:
## Describe the Bug
Deploying new deployment of `eks/cluster` will fail due to the OIDC not existing. The `local.eks_cluster_oidc_issuer_url` will initial be an empty string. However, the `cloudposse/eks-iam-role/aws` module for `aws-ebs-csi-driver` and `vpc-cni` addons requires the variable even if the addons are _not_ enabled.

## Expected Behavior
The `eks/cluster` component should work for initial deployments.

## Steps to Reproduce
1. deploy a **new** `eks/cluster` using the same snippet in the readme

## Screenshots
n/a

## Environment (please complete the following information):
 - OS: linux
 - latest terraform
 - latest aws terraform provider

## Additional Context
n/a

PR #798, opened 2023-08-07
bug: eks cluster fix first deploy due to OIDC

sgtoj:
## what
- set a non-empty value for `eks_cluster_identity_oidc_issuer` as the `default` for `module.eks` (remote-state module). 

## why
- that value must not be empty when passed to the `cloudposse/eks-iam-role/aws` module for `aws-ebs-csi-driver` and `vpc-cni` addons
  - that will never exists on first deployment because the component itself create the OIDC provider

## references
- #797


sgtoj:
This fix was one of two or three methods I could have implemented as a fix. If this is rejected I can provide another solution.

modules/eks/cluster/remote-state.tf
```diff
@@ -48,7 +48,7 @@ module "eks" {
     eks_managed_node_workers_role_arns = []
     fargate_profile_role_arns          = []
     fargate_profile_role_names         = []
-    eks_cluster_identity_oidc_issuer   = ""
+    eks_cluster_identity_oidc_issuer   = "https://placeholder.example.com/id/ABCDEF00000000000000000000000000"
```
milldr, line 51:
```suggestion
    # eks_cluster_identity_oidc_issuer will never exist on the first deployment because the component itself creates the OIDC provider, 
    # yet the value must not be empty when passed to the aws-ebs-csi-driver and vpc-cni addons
    eks_cluster_identity_oidc_issuer   = "https://placeholder.example.com/id/ABCDEF00000000000000000000000000" 
```

Explanation in comments for the placeholder

Nuru (changes requested):
This should not be necessary. The Add-Ons are not created until after cluster, at which point the OIDC issuer is available. I have not seen any problems creating clusters.

modules/eks/cluster/remote-state.tf
```diff
@@ -48,7 +48,9 @@ module "eks" {
     eks_managed_node_workers_role_arns = []
     fargate_profile_role_arns          = []
     fargate_profile_role_names         = []
-    eks_cluster_identity_oidc_issuer   = ""
+    # eks_cluster_identity_oidc_issuer will never exist on the first deployment because the component itself creates the OIDC provider,
+    # yet the value must not be empty when passed to the aws-ebs-csi-driver and vpc-cni addons
+    eks_cluster_identity_oidc_issuer = "https://placeholder.example.com/id/ABCDEF00000000000000000000000000"
```
Nuru, line 53: This would not fix the problem, it would only hide it. The add-on will still not have the OIDC issuer it needs, but with this it will accept the bad value and defer the error to logs after it is deployed.

PR #799, opened 2023-08-07
feat: allow email to be configured at account level

sgtoj:
## what
* allow email to be configured at account level

## why
* to allow importing existing accounts with email address that does not met the organization standard naming format  

## references
* n/a

### DIFF
--- a/modules/account/main.tf
+++ b/modules/account/main.tf
@@ -13,7 +13,7 @@ locals {
   # Organizational Units' Accounts list and map configuration
   organizational_units_accounts = flatten([
     for ou in local.organizational_units : [
-      for account in lookup(ou, "accounts", []) : merge(account, { "ou" = ou.name, "account_email_format" = lookup(ou, "account_email_format", var.account_email_format) })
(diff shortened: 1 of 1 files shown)


modules/account/main.tf
```diff
@@ -13,7 +13,7 @@ locals {
   # Organizational Units' Accounts list and map configuration
   organizational_units_accounts = flatten([
     for ou in local.organizational_units : [
-      for account in lookup(ou, "accounts", []) : merge(account, { "ou" = ou.name, "account_email_format" = lookup(ou, "account_email_format", var.account_email_format) })
+      for account in lookup(ou, "accounts", []) : merge(account, { "ou" = ou.name, "account_email_format" = lookup(account, "account_email_format", lookup(ou, "account_email_format", var.account_email_format)) })
```
Nuru, line 16: Just change this so that explicit values in the account take precedence over inherited or default values.

```suggestion
      for account in lookup(ou, "accounts", []) : merge({ "ou" = ou.name, "account_email_format" = lookup(ou, "account_email_format", var.account_email_format) }, account)
```

Nuru (changes requested):

Nuru (approved):

PR #802, opened 2023-08-07
bug: handle lowercase name tag when explicitly defining Name tag

sgtoj:
## what
* gracefully handle lowercase `name` tag when explicitly defining `Name` tag
  - setting `name` to `null` means the provider ignore the key-value pair and therefore creates no conflicts with `Name`

## why
* aws provider (least with latest provider) does not allow both `name` and `Name` keys to exists for `tags` variable

## references
* n/a

### DIFF
--- a/modules/aws-team-roles/main.tf
+++ b/modules/aws-team-roles/main.tf
@@ -69,7 +69,7 @@ resource "aws_iam_role" "default" {
   description          = each.value["role_description"]
   assume_role_policy   = data.aws_iam_policy_document.assume_role_aggregated[each.key].json
   max_session_duration = each.value["max_session_duration"]
-  tags                 = merge(module.this.tags, { Name = local.role_name_map[each.key] })
+  tags                 = merge(module.this.tags, { Name = local.role_name_map[each.key], name = null })
 }
 
 resource "aws_iam_role_policy_attachment" "default" {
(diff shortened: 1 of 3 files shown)


Nuru (changes requested):
@sgtoj Can you please explain this use case? This solution only handles the `name` tag, but you would have the same problem if you included other tags where the keys were distinct under case-sensitive matches but identical under case-insensitive matches. So unless we have a comprehensive solution, I'm inclined to leave it to users to clean up their inputs.

sgtoj:
> So unless we have a comprehensive solution, I'm inclined to leave it to users to clean up their inputs.

Yes. Whenever `label_key_case` is set to lower in the global stack file, then both name and `Name` will exists in the situations this issue resolves. 

I also realize sometimes `Name` is intentional even if all other tags are normalized to lowercase because the AWS web console will the value of `Name` as the name if it exists.

The solution addresses `Name`/`name` becuase that's the only tag explicitly being defined and does so without using processing through `cloudposse/label/null` module (so it can be normalized per configuration).

Nuru:
I am closing this as "wontfix" because it necessarily adds warts to the code without solving the underlying problem. @sgtoj Feel free to [open an issue on null-label](https://github.com/cloudposse/terraform-null-label/issues) about filtering out conflicting tags.

sgtoj:
@Nuru - I dont think this solves the problem outlined in the OP that exists now this repo's current state. The 3 components cannot be deployed as is if `label_key_case` is set to `lower`. At the very least, the tags should be normalized in via null-label instead of explicitly setting `Name`. 

Opening in issue on the null-label repo is a good idea nevertheless, but for different reasons.

sgtoj:
@Nuru - I have no issue submitting an updated PR that normalizes the tags as I suggested.

PR #804, opened 2023-08-08
Transit Gateway Cross-Region Support

milldr:
## what
- Upgraded `tgw` components to support cross region connections
- Added back `tgw/cross-region-hub-connector` with overhaul to support updated `tgw/hub` component

## why
- Deploy `tgw/cross-region-hub-connector` to create peered TGW hubs
- Use `tgw/hub` both for in region and intra region connections

## references
- n/a


max-lobur (approved):

Benbentwo (approved):
looks good, glad we brought back cross-region, only have 1 spoke component clears up understanding too

PR #809, opened 2023-08-09
Update EC2-Autoscale-Group Modules to 0.35.1

RoseSecurity:
## What:

- Updated `modules/spacelift/worker-pool` from 0.34.2 to 0.35.1 and adapted new variable features
- Updated `modules/bastion` from 0.35.0 to 0.35.1 
- Updated `modules/github-runners` from 0.35.0 to 0.35.1 

## Why:

- Modules were utilizing previous `ec2-autoscale-group` versions

## References:

- [terraform-aws-ec2-autoscale-group](https://github.com/cloudposse/terraform-aws-ec2-autoscale-group/blob/main/variables.tf)
- [Terraform Registry](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#instance_refresh)


modules/spacelift/worker-pool/variables.tf
```diff
@@ -215,6 +215,8 @@ variable "instance_refresh" {
     preferences = object({
       instance_warmup        = number
       min_healthy_percentage = number
+      skip_matching          = bool
+      auto_rollback          = bool
```
milldr, line 219: Can we set these all as optional?

```
      instance_warmup        = optional(number, null)
      min_healthy_percentage = optional(number, null)
      skip_matching          = optional(bool, null)
      auto_rollback          = optional(bool, null)
```

milldr: https://github.com/cloudposse/terraform-aws-ec2-autoscale-group#input_instance_refresh

RoseSecurity: For sure! Just pushed those changes. Great idea with that as well

milldr (approved):

PR #831, opened 2023-08-22
chore: remove defaults from components

dudymas:
## what
* remove `defaults.auto.tfvars` from component modules

## why
* in favor of drying up configuration using atmos

## Notes
* Some defaults may not be captured yet. Regressions might occur.


modules/eks/aws-node-termination-handler/default.auto.tfvars
Benbentwo, line 1: this file either needs to be reverted, or modified vars, as many of these are indeed sane defaults. it doesn't look like they're in the readme either

modules/eks/metrics-server/default.auto.tfvars
Benbentwo, line 1: same as above

modules/eks/aws-node-termination-handler/default.auto.tfvars
Benbentwo, line 1: These should be moved to var defaults or the readme

modules/eks/metrics-server/default.auto.tfvars
Benbentwo, line 1: same as above

modules/zscaler/default.auto.tfvars
```diff
@@ -1,6 +0,0 @@
-enabled = false
-
-name = "zscaler"
-
-# Cheapest instance that satisfies DenyInstancesWithoutEncryptionInTransit SCP (see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/data-protection.html#encryption-transit)
-instance_type = "m5n.large"
```
Benbentwo, line 6: variable default

Benbentwo (changes requested):

Benbentwo (approved):

PR #836, opened 2023-08-24
Update helm release for eks/external-dns

mfuhrmeisterDM:
## what
* upgrade helm release
* updating also from deprecated `iam_policy_statements` to `iam_source_policy_documents`


Nuru (changes requested):
Since you opened this PR, we have updated this module to use `helm-release` v0.9.1, which is now causing a merge conflict. If that is good enough for you, you can just close this PR and we will update this module to use v0.9.3 later.

If you want to continue with this PR:

Please resolve the merge conflict (you can just rebase your changes on `main`).

Then, you need to run the pre-commit hooks. (If you have already committed all your changes as you would have to in order to rebase, un-commit them with `git reset --soft HEAD~1` first.)

```
make init
git add -u && make pre-commit/run
```

Then commit everything and push (`--force-with-lease` if necessary).

PR #843, opened 2023-08-27
Update helm-release to 0.9.3 for eks/karpenter

mgledi:
## what

- Update helm-release version from 0.7.0 to 0.9.4

## why

- get rid of deprecated old iam_policy_statements


Nuru (changes requested):
Same requests as for your [eks/external-dns PR #836](https://github.com/cloudposse/terraform-aws-components/pull/836#pullrequestreview-1598896987)


Please also add to the comment regarding the policy that the official version of the policy is in the Cloud Formation Template at https://github.com/aws/karpenter/blob/61cc8f75049eafbff4011147f8e272fe18c91157/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml#L41-L221

but cannot be used directly because it includes Cloud Formation variables that need to be substituted.

PR #847, opened 2023-08-30
TGW Spoke Association and Propagation

milldr:
## what
- Return the transit gateway association and propagation creation to all `tgw/spoke` deployments

## why
- We conditional set the creation for alternate VPCs in the same region. However, all VPCs do need a separate transit gateway association and propagation in the network account, which is what this setting is enabling.

## references
- n/a


mcalhoun (approved):
LGTM! (already reviewed the downstream version in the customer repo).

PR #849, opened 2023-08-30
feat: spacelift/worker-pool can fully name and re-use iam role

dudymas:
## what

- aws-teams supports setting `allowed_roles` for a team, adding trust to roles in other accounts.
- add version var to `spacelift/worker-pool` to allow refreshing instances
- update remote-state in `spacelift/worker-pool` to make sure `enabled=false` doesn't break terraform.

## why

- better support for tfstate permissions
- allows configuring spacelift in teams as follows:
```
          spacelift:
            <<: *user-template
            role_description: Team for our privileged Spacelift server
            role_policy_arns:
            - team_role_access
            aws_saml_login_enabled: false
            allowed_roles:
              core-auto:
                - spacelift-worker-pool
            trusted_teams:
              - devops
              - managers
            trusted_role_arns: []
```


modules/spacelift/worker-pool/remote-state.tf
```diff
@@ -36,9 +36,9 @@ module "spaces" {
   version = "1.5.0"
 
   component   = var.spacelift_spaces_component_name
-  environment = try(var.spacelift_spaces_environment_name, module.this.environment)
-  stage       = try(var.spacelift_spaces_stage_name, module.this.stage)
-  tenant      = try(var.spacelift_spaces_tenant_name, module.this.tenant)
+  environment = try(var.spacelift_spaces_environment_name, var.environment)
+  stage       = try(var.spacelift_spaces_stage_name, var.stage)
+  tenant      = try(var.spacelift_spaces_tenant_name, var.tenant)
```
milldr, line 41: I could be missing the intention, but we should not need this or any of the following changes from `module.this.tenant/stage/environment` to `var.tenant/stage/environment`

The `spacelift/worker-pool` component has a given tenant, stage, and environment. `module.this` will take those values as default. 

However, since `spacelift/worker-pool` is deployed in `core-use2-spacelift`, we need to specify a different tenant, stage, and environment for where to find `account-map`. That's the purpose of `var.account_map_tenant_name`, `var.account_map_stage_name`, and `var.account_map_environment_name` Similarly, `spacelift/spaces` has a unique stack. We also already have variables to specify where that is: `var.spacelift_spaces_{tenant/stage/environment}_name`.

What's the purpose of these new variables that `var.account_map_{tenant/stage/environment}_name` and/or `var.spacelift_spaces_{tenant/stage/environment}_name` do not accomplish?

milldr (changes requested):

modules/spacelift/worker-pool/variables.tf
```diff
@@ -202,14 +202,49 @@ variable "spacelift_domain_name" {
   default     = "spacelift.io"
 }
 
+variable "create_role" {
+  type        = bool
+  description = "Whether to create an IAM role for the worker pool"
```
milldr, line 207:
```suggestion
  description = "Whether or not to create an IAM role for the worker pool"
```

modules/spacelift/worker-pool/iam.tf
```diff
@@ -4,7 +4,7 @@ module "iam_label" {
 
   attributes = var.iam_attributes
 
-  context = module.this.context
+  context = merge(module.this.context, var.iam_context)
```
milldr, line 7:
```suggestion
  tenant = var.teams_tenant_name
  stage = var.teams_stage_name
  environment = var.teams_environment_name
  
  context = module.this.context
```

This is a more opinionated suggestion than the other. The other context variables for `account-map` or `spacelift/spaces` are changed by an individual variable, rather than a merged object. IMO this should be consistent so that it's easier to understand

Set the default for each to `null`. Null label will take the value from context if not defined.

milldr: However taking a step back, why do we need to change these context options?

modules/spacelift/worker-pool/remote-state.tf
dudymas: If you set `enabled=false`, the component will not tear down. The `module.this` causes errors. I can get an example going, but it was done to better support `enabled=false`

modules/spacelift/worker-pool/iam.tf
dudymas: Aha, yeah, I should have explained this. In the end, I was looking at errors like this one:
```
An error occurred (AccessDenied) when calling the AssumeRole operation: User: 
arn:aws:sts::123456789012:assumed-role/acme-core-gbl-auto-spacelift-worker-pool/i-12345678901234567

is not authorized to perform: sts:AssumeRole on resource: 
arn:aws:iam::123456789013:role/acme-core-gbl-root-tfstate
```

dudymas: When update the allowed roles in `tfstate-backend`, they are created with `gbl` as their environment. Given the choice between the two, it felt easier and less disruptive to allow the iam-role made in `spacelift/worker-pool` to change than to update `tfstate-backend` to allow roles that have something other than `gbl` in their role name.

milldr (approved):

PR #853, opened 2023-09-01
ref: remove allowed_roles from teams_config

dudymas:
## what

- remove `allowed_roles` from `teams_config` in `aws-teams`

## why

- roles shouldn't be allowed to assume teams. And exceptions must be specified
in trusted arns list


Nuru:
One can debate the usefulness of this feature, or argue that it is a bad idea to allow a role to assume a team, but I see nothing inherently wrong about having this feature available, and it is possible someone is using it, so I am rejecting this PR as `wontfix`.

PR #862, opened 2023-09-21
Update README.md

noeljackson:
Include examples for numbers, and long strings, such as DKIM records.

## what

examples for records starting with numbers, or DKIM records with long values

## why

We faced issues with the existing examples


milldr (approved):
LGTM! thanks for updating these docs

PR #871, opened 2023-10-12
chore: add instance_refresh, al2023 to github-runners

dudymas:
## what

- add `instance_refresh` and `launch_template_version` to `github-runners`
- support al2023
- remove `ecr-login` configuration from user-data.sh

## why

- make it less click-ops to update github runner versions
- node20 requires newer glibc, so updating to al2023 and installing `libicu` solves that requirement
- the `ecr-login` configuration does not support docker login due to [PR 315](https://github.com/awslabs/amazon-ecr-credential-helper/pull/315)
- the `ecr-login` configuration cannot use [environment credential
sources](https://github.com/awslabs/amazon-ecr-credential-helper/issues/181)
- README has been updated for `github-runners` to explain the most flexible
usage of this component with ECR


milldr (dismissed):

Nuru (dismissed):
Do not ignore launch template version. If `instance_refresh` is true and `launch_template_version = "$Latest"`, configure `launch_template_version` to use the `latest_version` attribute of the `aws_launch_template` resource.

Add the ability to set `instance_refresh.triggers`.

Set appropriate options for instance refresh for GitHub Runners, such as `scale_in_protected_instances = "Refresh"`.

dudymas:
This looks like it's too stale and needs to be closed. Also, the runs-on component is a better fit for modern infra.

PR #875, opened 2023-10-13
Variable names for `aurora-mysql`

milldr:
## what
- Corrected variable names for the `aurora-mysql` component

## why
- `var.cluster_size` is used in `aurora-postgres`, not `aurora-mysql`. This should be `var.mysql_cluster_size`
- The Readme referred to incorrect variable names

## references
- JBI-153


bradj (approved):
👍

PR #879, opened 2023-10-18
`datadog-synthetics-private-location` Kubernetes Namespace

milldr:
## what
- Fixed namespace issues with datadog-synthetics-private-location
- Added a comment

## why
- The input for `helm-release` for a namespace is `kubernetes_namespace`, not `namespace`
- Comment to explain that namespace is usually created with the `eks/datadog-agent` component, so we usually don't need to create it here.

## references
- https://github.com/cloudposse/terraform-aws-helm-release#input_kubernetes_namespace


bradj (approved):
👍

PR #880, opened 2023-10-18
Support for Multi Region Clusters for ArgoCD

milldr:
## what
- Added environment to `saml_sso_providers` input for `eks/argocd`

## why
- The `saml_sso_providers` is used to read the output of another component. The SSO implementation of that component requires manually adding a few SSM parameters, but those values can be reused across regions. Therefore we only need to deploy `saml_sso_provider` once per stage, and then specify which environment to read from (if in an alternate region).

## references
- n/a


bradj (approved):
👍

PR #881, opened 2023-10-20
Allow VPC Ingress from Multiple Regions in Same Stage for `eks/cluster`

milldr:
## what
- Include `environment` in map key when iterating through `var.allow_ingress_from_vpc_accounts` for `eks/cluster`

## why
- VPC accounts with the same `stage` and (optional) `tenant` but different `environments` in `var.allow_ingress_from_vpc_accounts` will otherwise have the same map key, and Terraform will fail

## references
- n/a


bradj (approved):
👍

PR #882, opened 2023-10-20
`eks/external-secrets-operator` Overridable Additional Policy Statements

milldr:
## what
- Update module versions
- Add optional override for additional policies for `eks/external-secrets-operator`

## why
- We may want to grant additional access to the operator. With the override pattern, we can grant whatever we need for specific use-case and avoid conflicts with upstream

## references
- override pattern reference: https://github.com/cloudposse/terraform-aws-components/blob/main/modules/aws-teams/additional-policy-map.tf


bradj (approved):
👍

PR #890, opened 2023-11-01
fix: aws-saml/okta-user should handle enable flag

dudymas:
## what

- make sure references to potentially disabled resources are handled

## why

- aws-saml/okta-user was erroring out on references that didn't handle the
`count` updates from #805


Nuru (changes requested):
If the module is not enabled, it should not create any resources (or use data sources). Fix `aws_iam_policy` etc.

hans-d:
@dudymas seems to be fixed with #903

dudymas:
Closing as another PR fixed this

PR #892, opened 2023-11-01
Passthough AMI ID list to dynamic-subnets module

mfuhrmeisterDM:
## why
We want to be able to specify a list of ami ids to use for the nat instance(s)

## what
passthought that list, default is empty


mfuhrmeisterDM:
> @mfuhrmeisterDM thank you for the PR, please address the comment

its applied

PR #897, opened 2023-11-06
aws-shield is now able to protect alb's from ingress-groups

bradj:
## what

- allows `aws-shield` to pull alb name from eks ingress group

## why

- simpler `aws-shield` provisioning


modules/aws-shield/alb.tf
```diff
@@ -1,5 +1,5 @@
 data "aws_alb" "alb" {
-  for_each = local.alb_protection_enabled ? toset(var.alb_names) : []
+  for_each = local.alb_protection_enabled == false ? [] : length(var.alb_names) > 0 ? toset(var.alb_names) : toset([module.alb.outputs.load_balancer_name])
```
milldr, line 2:
```suggestion
  for_each = local.alb_protection_enabled == false ? [] : length(var.alb_names) > 0 ? toset(var.alb_names) : toset([module.alb[0].outputs.load_balancer_name])
```

milldr (changes requested):
we can get the alb names from remote state by default if `alb_protection_enabled` is `true` and not need another variable. If `var.alb_names` is set, then instead use those names

We'll probably also need to update the local definition for `local.alb_protection_enabled` to use that option

modules/aws-shield/remote-state.tf
```diff
@@ -0,0 +1,9 @@
+module "alb" {
+  count   = length(var.alb_names) > 0 ? 0 : 1
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "1.5.0"
+
+  component = "eks/alb-controller-ingress-group"
```
milldr, line 6: nitpick: `eks` component name changes sometimes. this should be a variable but probably isnt required right now

milldr (approved):

PR #903, opened 2023-11-07
Bugfix: Okta Submodule in `aws-saml`

Benbentwo:
## what

 - Bugfix: `aws-saml` component without okta giving issue about not picking a member of the array since enabled flag was added with count in #805 

## why

- deploy component whether or not okta is used

## references

 - #805


hans-d:
Aslo #890

milldr (approved):

PR #907, opened 2023-11-14
Fix fetching setting for plat and dev spacelift stacks

mfuhrmeisterDM:
## why
for the plat and core spacelift modules the variable `root_admin_stack_config` is not set. Therefore the stack config null, but in our case we need `github_enterprise` to be set.

## what
Read the config from the stack config as it is there also available.


mfuhrmeisterDM:
closing in favour of #904

PR #909, opened 2023-11-20
New Component: `philips-labs-github-runners`

Benbentwo:
## what

 - The [philips-labs/terraform-aws-github-runner](https://github.com/philips-labs/terraform-aws-github-runner) is a popular way to deploy scalable self-hosted GitHub Runners on AWS. This PR introduces a component to deploy those runners.
 - This uses their default example as the basis of this component.

## why

 - When not using EKS and actions runner controller, we need a scalable way to use self-hosted GitHub runners. Currently, we use an ASG-based component. This is slower and less responsive to surges, furthermore, it scales down slower, resulting in higher operating costs.

## references

 * https://github.com/philips-labs/terraform-aws-github-runner
 
## Future Improvements
 - The Philips lab module currently has support for yaml based multi-arch runners. This would fit nicely into the Cloud Posse Reference architecture with stacks and atmos merging.

### DIFF
--- a/modules/philips-labs-github-runners/README.md
+++ b/modules/philips-labs-github-runners/README.md
@@ -1,0 +1,141 @@
+# Component: `philips-labs-github-runners`
+
+This component is responsible for provisioning the surrounding infrastructure for the github runners.
+
+## Prerequisites
+
+* Github App installed on the organization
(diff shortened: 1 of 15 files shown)


dudymas (dismissed):
LGTM! Thanks Ben

PR #914, opened 2023-11-30
fix(inspector is now enabled on deploy)

bradj:
## what

- enables Inspector on Delegated Admin and Member accounts

## why

- deployments would fail due to `invoking account not being enabled`


mcalhoun (approved):
LGTM

PR #928, opened 2023-12-11
feat: create `aws-sso` groups for google workspaces

milldr:
## what
- create Identity Center Groups with an optional `var.groups` in `aws-sso`

## why
- Identity Center + google now is supported, but groups are not automatically created as they would be with Okta or Jumpcloud for example
- Google will automatically sync users into groups, but the groups must first exist

## references
- https://docs.aws.amazon.com/singlesignon/latest/userguide/gs-gwp.html


modules/aws-sso/variables.tf
```diff
@@ -42,3 +42,13 @@ variable "aws_teams_accessible" {
     EOT
   default     = []
 }
+
+variable "groups" {
+  type        = list(string)
+  description = <<-EOT
+    List of AWS Identity Center Groups to be created with the AWS API.
+
+    When provisioning the GSuite Integration with AWS, Groups need to be created with API in order for automatic provisioing to work as intended.
```
dudymas, line 51:
```suggestion
    When provisioning the Google Workspace Integration with AWS, Groups need to be created with API in order for automatic provisioning to work as intended.
```

modules/aws-sso/outputs.tf
```diff
@@ -7,3 +7,8 @@ output "sso_account_assignments" {
   value       = module.sso_account_assignments.assignments
   description = "SSO account assignments"
 }
+
+output "group_ids" {
+  value       = { for group_key, group_output in aws_identitystore_group.manual : group_key => group_output.group_id }
+  description = "Group IDs for created in Identity Center"
```
dudymas, line 13:
```suggestion
  description = "Group IDs created for Identity Center"
```

dudymas (approved):
one more verbiage change I'd recommend for outputs... otherwise looks great!

dudymas (approved):
LGTM
image

PR #934, opened 2023-12-14
[`Draft`] Component: `aws-sso-app`

Benbentwo:
## what

* initial component of `aws-sso-app` for visibility

## why

* we recommend using AWS Identity Center applications for signing into various applications to simplify the sign in experience.

## references
**Requires**
* https://github.com/hashicorp/terraform-provider-aws/issues/27853
* https://github.com/hashicorp/terraform-provider-aws/issues/34813
* https://github.com/aws/aws-sdk/issues/747


modules/aws-sso-app/README.md
milldr, line 1: we should avoid using "aws sso" as a term for new work since it's now called "identity center". Although it would be confusing with the existing `aws-sso` component

Benbentwo:
Closing this as it will not be completed (due to required dependent GH Issues listed above) by the time we migrate our components to our new org. 

I will consider reopening this at a later time in a new component repository

PR #936, opened 2023-12-19
`github-oidc-role` component

Benbentwo:
## what

 - Component for creating GitHub OIDC roles - allows GHA to assume the role and interact with AWS Services

## why

Currently to add a policy, we recommend adding a `github-actions-iam-policy.tf` file with your policy defined, then vendoring in [the mixin `github-actions-iam-role.mixin.tf`](https://github.com/cloudposse/terraform-aws-components/blob/main/mixins/github-actions-iam-role/github-actions-iam-role.mixin.tf) to deploy the policy.

This works well, but sometimes a policy doesn't fit in a particular component, such as needing access to S3 and SSM. We could always add it in, but it makes it unclear why it is coupled so tightly with a component.

This component allows the creation of GitHub OIDC roles via instances of this component. We intend to add more policies to the component as needed but also allow AWS-managed policies, and stack inline defined policies which can be used and mixed as needed.

## references

 - closes: #930


modules/github-oidc-role/README.md
```diff
@@ -0,0 +1,190 @@
...
+This component is responsible for creating IAM roles for GitHub Actions to assume.
+
+## Usage
+
+**Stack Level**: Regional
+
+Here's an example snippet for how to use this component.
+
+```yaml
+# stacks/catalog/gha-service-roles/defaults.yaml
```
milldr, line 12: replace `gha-service-roles` with `github-oidc-role`. plus other references in this README

modules/github-oidc-role/README.md
```diff
@@ -0,0 +1,190 @@
...
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| <a name="output_github_actions_iam_role_arn"></a> [github\_actions\_iam\_role\_arn](#output\_github\_actions\_iam\_role\_arn) | ARN of IAM role for GitHub Actions |
+| <a name="output_github_actions_iam_role_name"></a> [github\_actions\_iam\_role\_name](#output\_github\_actions\_iam\_role\_name) | Name of IAM role for GitHub Actions |
+<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+
+## References
+* [cloudposse/terraform-aws-components](https://github.com/cloudposse/terraform-aws-components/tree/master/modules/gha-service-roles) - Cloud Posse's upstream component
```
milldr, line 188:
```suggestion
* [cloudposse/terraform-aws-components](https://github.com/cloudposse/terraform-aws-components/tree/master/modules/github-oidc-role) - Cloud Posse's upstream component
```

modules/github-oidc-role/policy_gitops.tf
```diff
@@ -0,0 +1,107 @@
...
+
+  s3_bucket_arn      = one(module.s3_bucket[*].outputs.bucket_arn)
+  dynamodb_table_arn = one(module.dynamodb[*].outputs.table_arn)
+}
+
+module "s3_bucket" {
+  count = local.gitops_policy_enabled ? 1 : 0
+
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "1.4.3"
```
milldr, line 25: bump to latest

milldr: same for other remote-state modules

modules/github-oidc-role/providers.tf
milldr, line 1: use latest providers.tf
https://github.com/cloudposse/terraform-aws-components/blob/main/mixins/providers.depth-1.tf

modules/github-oidc-role/policy_lambda-cicd.tf
milldr, line 1: filename has both `-` and `_`. Should we use one or the other?

milldr: imo they should all use `-`. `policy_gitops.tf` should be renamed as well. that's consistent with how `aws-sso` names files

modules/github-oidc-role/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  policy_document_map = {
+    "gitops"        = local.gitops_policy
+    "lambda_cicd"   = local.lambda_cicd_policy
+    "inline_policy" = one(module.iam_policy.*.json)
+  }
+  active_policy_map = { for k, v in local.policy_document_map : k => v if v != null }
+}
+
+module "iam_policy" {
+  count = local.enabled && length(var.iam_policy) > 0 ? 1 : 0
```
milldr, line 14: since this is a cloud posse module, can we use `enabled` instead of `count`?

```suggestion
  enabled = local.enabled && length(var.iam_policy) > 0
```

modules/github-oidc-role/main.tf
```diff
@@ -0,0 +1,54 @@
+locals {
+  enabled             = module.this.enabled
+  aws_canned_policies = [for arn in var.aws_iam_policies : arn if can(regex("^arn:aws:iam::aws:policy/", arn))]
+  aws_policies        = length(local.aws_canned_policies) > 0 ? local.aws_canned_policies : null
+  policy_document_map = {
+    "gitops"        = local.gitops_policy
+    "lambda_cicd"   = local.lambda_cicd_policy
+    "inline_policy" = one(module.iam_policy.*.json)
+  }
```
Nuru, line 9: This needs to be overridable/extensible. See example from `aws-team-roles`: https://github.com/cloudposse/terraform-aws-components/blob/1b338fe664e5debc5bbac30cfe42003f7458575a/modules/aws-team-roles/main.tf#L10-L16

modules/github-oidc-role/main.tf
```diff
@@ -0,0 +1,54 @@
...
+}
+
+module "iam_policy" {
+  count = local.enabled && length(var.iam_policy) > 0 ? 1 : 0
+
+  source  = "cloudposse/iam-policy/aws"
+  version = "2.0.1"
+
+  iam_policy = var.iam_policy
+}
```
Nuru, line 20:
```suggestion

  context = module.this.context
}
```

modules/github-oidc-role/main.tf
```diff
@@ -0,0 +1,54 @@
...
+
+module "gha_role_name" {
+  source  = "cloudposse/label/null"
+  version = "0.25.0"
+
+  attributes = compact(concat(var.github_actions_iam_role_attributes, ["gha"]))
```
Nuru, line 26: If we are going to allow the user to supply a set of attributes just for the role, then we should not be forcing any on them.
```suggestion
  attributes = var.github_actions_iam_role_attributes
```

modules/github-oidc-role/main.tf
```diff
@@ -0,0 +1,54 @@
+locals {
+  enabled             = module.this.enabled
+  aws_canned_policies = [for arn in var.aws_iam_policies : arn if can(regex("^arn:aws:iam::aws:policy/", arn))]
+  aws_policies        = length(local.aws_canned_policies) > 0 ? local.aws_canned_policies : null
```
Nuru, line 4: The "aws" prefix is redundant, and given that the policies can be our custom names, somewhat misleading. Also, as long as you are using a regex, you might as well allow for other partitions.

```suggestion
  canned_policies = [for arn in var.iam_policies : arn if can(regex("^arn:aws[^:]*:iam::aws:policy/", arn))]
  policies        = length(local.canned_policies) > 0 ? local.canned_policies : null
```

If you prefer, you can replace "canned" with "aws_managed", which aligns with the AWS documentation.

modules/github-oidc-role/policy_gitops.tf
```diff
@@ -0,0 +1,107 @@
+variable "gitops_policy_configuration" {
+  type        = map(string)
+  default     = {}
+  description = <<-EOT
+    Configuration for the GitOps IAM Policy, valid keys are
+     - `s3_bucket_component_name` - Component Name of where to store the TF Plans in S3, defaults to `gitops/s3-bucket`
+     - `dynamodb_component_name` - Component Name of where to store the TF Plans in Dynamodb, defaults to `gitops/dynamodb`
+     - `s3_bucket_environment_name` - Environment name for the S3 Bucket, defaults to current environment
+     - `dynamodb_environment_name` - Environment name for the Dynamodb Table, defaults to current environment
```
Nuru, line 9: Make this an object with explicit defaults. Default to `null` for current environment.

modules/github-oidc-role/policy_gitops.tf
Nuru, line 25:
```suggestion
  version = "1.5.0"
```

modules/github-oidc-role/policy_gitops.tf
```diff
@@ -0,0 +1,107 @@
...
+  environment = lookup(var.gitops_policy_configuration, "s3_bucket_environment_name", module.this.environment)
+
+  context = module.this.context
+}
+
+module "dynamodb" {
+  count = local.gitops_policy_enabled ? 1 : 0
+
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "1.4.3"
```
Nuru, line 37:
```suggestion
  version = "1.5.0"
```

modules/github-oidc-role/policy_lambda-cicd.tf
```diff
@@ -0,0 +1,102 @@
...
+    Configuration for the lambda-cicd policy. The following keys are supported:
+      - `enable_kms_access` - (bool) - Whether to allow access to KMS. Defaults to false.
+      - `enable_ssm_access` - (bool) - Whether to allow access to SSM. Defaults to false.
+      - `enable_s3_access` - (bool) - Whether to allow access to S3. Defaults to false.
+      - `s3_bucket_component_name` - (string) - The name of the component to use for the S3 bucket. Defaults to `s3-bucket/github-action-artifacts`.
+      - `s3_bucket_environment_name` - (string) - The name of the environment to use for the S3 bucket. Defaults to the environment of the current module.
+      - `s3_bucket_tenant_name` - (string) - The name of the tenant to use for the S3 bucket. Defaults to the tenant of the current module.
+      - `s3_bucket_stage_name` - (string) - The name of the stage to use for the S3 bucket. Defaults to the stage of the current module.
+      - `enable_lambda_update` - (bool) - Whether to allow access to update lambda functions. Defaults to false.
+  EOT
```
Nuru, line 14: Make object with explicit defaults

modules/github-oidc-role/policy_lambda-cicd.tf
```diff
@@ -0,0 +1,102 @@
...
+  lambda_cicd_policy_enabled = contains(var.aws_iam_policies, "lambda-cicd")
+  lambda_cicd_policy         = local.lambda_cicd_policy_enabled ? one(data.aws_iam_policy_document.lambda_cicd_policy.*.json) : null
+
+}
+
+module "s3_artifacts_bucket" {
+  count = lookup(var.lambda_cicd_policy_configuration, "enable_s3_access", false) ? 1 : 0
+
+  source  = "cloudposse/stack-config/yaml//modules/remote-state"
+  version = "1.4.3"
```
Nuru, line 27:
```suggestion
  version = "1.5.0"
```

modules/github-oidc-role/policy_lambda-cicd.tf
```diff
@@ -0,0 +1,102 @@
...
+      effect = "Allow"
+      actions = [
+        "s3:HeadObject",
+        "s3:GetObject",
+        "s3:PutObject",
+        "s3:ListBucket",
+        "s3:GetBucketLocation"
+      ]
+      resources = [
+        module.s3_artifacts_bucket[0].outputs.bucket_arn,
```
Nuru, line 84: You cannot rely on `module.s3_artifacts_bucket[0].outputs.bucket_arn` being available, even though the logic suggests that this is required. During destruction, and particularly targeted destruction or final destruction after partial destruction succeeded but full destruction failed, you can have these "required" outputs be null.

Move `module.s3_artifacts_bucket[0].outputs.bucket_arn` to a local variable, using `try(..., null)`, then check if the value is null in the `for_each` statement.

modules/github-oidc-role/variables.tf
```diff
@@ -0,0 +1,70 @@
+variable "region" {
+  type        = string
+  description = "AWS Region"
+}
+
+variable "aws_iam_policies" {
```
Nuru, line 6:
```suggestion
variable "iam_policies" {
```

modules/github-oidc-role/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+
+
+variable "github_actions_iam_role_enabled" {
+  type        = bool
+  description = <<-EOF
+  Flag to toggle creation of an IAM Role that GitHub Actions can assume to access AWS resources
+  EOF
+  default     = false
+}
+
```
Nuru, line 55: This seems redundant. When would you enable the component but leave the role creation disabled? I suggest removing it.

```suggestion
```

modules/github-oidc-role/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+  type        = list(string)
+  description = <<EOF
+  A list of the GitHub repositories that are allowed to assume this role from GitHub Actions. For example,
+  ["cloudposse/infra-live"]. Can contain "*" as wildcard.
+  If org part of repo name is omitted, "cloudposse" will be assumed.
+  EOF
+  default     = []
+}
+
+variable "github_actions_iam_role_attributes" {
```
Nuru, line 66: I am uncomfortable with this, because all this module does is create a role, so why isn't the standard `null-label` attributes enough? But given that attributes are lists and lists do not get merged, I will let it go.

modules/github-oidc-role/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.0.0"
```
Nuru, line 2: Use optional object members and require Terraform 1.3.0 to support that.
```suggestion
  required_version = ">= 1.3.0"
```

Nuru (changes requested):

modules/github-oidc-role/policy_lambda-cicd.tf
Nuru: I'm in favor of using the underscore because it distinguished `policy` as a prefix from the role name, which can have hyphens. 

That said, `aws-teams` uses `policy-...` so I guess I lose on that one.

Benbentwo: I'm in favor of underscore separated and hypens for the name. this is because in stack config when stating which policy you want to use, we usually use hyphens. however we do not include `policy`.

Similar to what @Nuru said

milldr (dismissed):

modules/github-oidc-role/README.md
```diff
@@ -37,19 +37,22 @@ components:
...
           - "MyOrg/infrastructure"
-        aws_iam_policies:
+        attributes: [ "gitops" ]
+        iam_policies:
           - gitops
         gitops_policy_configuration:
           s3_bucket_component_name: gitops/s3-bucket
           dynamodb_component_name: gitops/dynamodb
+          s3_bucket_environment_name: use2 # replace with your environment
+          dynamodb_environment_name: use2 # replace with your environment
```
Nuru, line 49: Shouldn't these be left to default?

modules/github-oidc-role/additional-policy-map.tf
```diff
@@ -0,0 +1,10 @@
+locals {
+  # If you have custom policies, override this declaration by creating
+  # a file called `additional-policy-map_override.tf`.
+  # Then add the custom policies to the overridable_additional_custom_policy_map in that file.
+  # See the README in `aws-team-roles` for more details.
```
Nuru, line 5: Explain this in the README in this component

modules/github-oidc-role/additional-policy-map.tf
```diff
@@ -0,0 +1,10 @@
...
+  overridable_additional_custom_policy_map = {
+    # Example:
+    #   eks_viewer = aws_iam_policy.eks_viewer.arn
```
Nuru, line 8: Clarify that this component, unlike others, takes a policy definition in JSON, not a customer managed policy ARN.

modules/github-oidc-role/main.tf
```diff
@@ -1,29 +1,24 @@
...
+  policies         = length(local.managed_policies) > 0 ? local.managed_policies : null
   policy_document_map = {
     "gitops"        = local.gitops_policy
     "lambda_cicd"   = local.lambda_cicd_policy
     "inline_policy" = one(module.iam_policy.*.json)
   }
-  active_policy_map = { for k, v in local.policy_document_map : k => v if v != null }
+  custom_policy_map = merge(local.policy_document_map, local.overridable_additional_custom_policy_map)
+
+  active_policy_map = { for k, v in local.custom_policy_map : k => v if v != null }
```
Nuru, line 12: We have seen "gotchas" where the policy ends up being empty instead of null, the empty policy being `"{}"`, so maybe
```suggestion
  # Ignore empty policies of the form `"{}"` as well as null policies
  active_policy_map = { for k, v in local.custom_policy_map : k => v if try(length(v), 0) > 3 }
```

modules/github-oidc-role/policy_lambda-cicd.tf
```diff
@@ -1,6 +1,15 @@
...
+  default = {
+    enable_kms_access          = "false"
+    enable_ssm_access          = "false"
+    enable_s3_access           = "false"
+    s3_bucket_component_name   = "s3-bucket/github-action-artifacts"
+    s3_bucket_environment_name = null
+    s3_bucket_tenant_name      = null
+    s3_bucket_stage_name       = null
+    enable_lambda_update       = "false"
+  }
```
Nuru, line 12:
```suggestion
  type     = object({
    enable_kms_access          = optional(bool, false)
    enable_ssm_access          = optional(bool, false)
    enable_s3_access           = optional(bool, false)
    s3_bucket_component_name   = optional(string, "s3-bucket/github-action-artifacts")
    s3_bucket_environment_name = optional(string)
    s3_bucket_tenant_name      = optional(string)
    s3_bucket_stage_name       = optional(string)
    enable_lambda_update       = optional(bool, false)
  }
  default  = {}
  nullable = false
```

modules/github-oidc-role/policy_gitops.tf
```diff
@@ -1,6 +1,11 @@
 variable "gitops_policy_configuration" {
-  type        = map(string)
-  default     = {}
+  type = map(string)
+  default = {
+    s3_bucket_component_name   = "gitops/s3-bucket"
+    dynamodb_component_name    = "gitops/dynamodb"
+    s3_bucket_environment_name = null
+    dynamodb_environment_name  = null
+  }
```
Nuru, line 8:
```suggestion
  type     = object({
    s3_bucket_component_name   = optional(string, "gitops/s3-bucket")
    dynamodb_component_name    = optional(string, "gitops/dynamodb")
    s3_bucket_environment_name = optional(string)
    dynamodb_environment_name  = optional(string)
  }
  default  = {}
  nullable = false
```

Nuru (changes requested):

modules/github-oidc-role/additional-policy-map.tf
```diff
@@ -2,9 +2,10 @@ locals {
   # If you have custom policies, override this declaration by creating
   # a file called `additional-policy-map_override.tf`.
   # Then add the custom policies to the overridable_additional_custom_policy_map in that file.
-  # See the README in `aws-team-roles` for more details.
+  # The key should be the policy you want to override, the value is the json policy document.
+  # See the README in `github-oidc-role` for more details.
```
Nuru, line 6: OK, but now you need to explain this in the README. You can use https://github.com/cloudposse/terraform-aws-components/blob/main/modules/aws-team-roles/README.md#privileges-are-granted-to-users-via-iam-policies as a starting point, but don't just copy and paste because the details are different.

Nuru (changes requested):
Please update README with instructions for adding custom policies

Nuru (changes requested):
The `policy_gitops` should be moved to its own component, because it creates an S3 bucket and DynamoDB table. This component should just create roles.

Nuru (approved):

PR #938, opened 2023-12-21
chore(ecs): Remove unused variables

hans-d:
## what

- remove unused vars from ecs component

## why

- not used


dudymas (approved):
lgtm

PR #945, opened 2024-01-06
feat: use `account-map` component for brownfield env

nitrocode:
## what

- feat: use `account-map` component for brownfield env

## why

- Allow brownfield environments to use the account-map component with existing accounts

## references

- Related to PR #943


modules/account-map/main.tf
```diff
@@ -6,10 +6,9 @@ locals {
   aws_partition               = data.aws_partition.current.partition
   legacy_terraform_uses_admin = coalesce(var.legacy_terraform_uses_admin, !var.terraform_dynamic_role_enabled)
 
-  full_account_map = {
-    for acct in data.aws_organizations_organization.organization.accounts
-    : acct.name == var.root_account_aws_name ? var.root_account_account_name : acct.name => acct.id if acct.status != "SUSPENDED"
-  }
+  non_suspended_account_ids = [for k, v in data.aws_organizations_organization.organization.accounts : v.id if v.status != "SUSPENDED"]
+
+  full_account_map = { for k, v in module.accounts.outputs.account_info_map : k => v.id if contains(local.non_suspended_account_ids, v.id) }
```
Nuru, line 11: What is the advantage of this change?

Note that this is unacceptable in any case because it removes the renaming of the root account.

Nuru (changes requested):

goruha:
Close this PR as it does not add any additional value

nitrocode:
Ouch

PR #949, opened 2024-01-09
update ecs cluster and services to support vanity domians

Benbentwo:
## what

update ecs cluster and services to support vanity domains

## why

vanity domain bugfix for new outputs from alb module and adding support for acm certs to be added to cluster alb's


dudymas (approved):
lgtm

PR #955, opened 2024-01-18
`ecs-service` additional_targets var

Benbentwo:
## what

 Currently `vanity_alias` allows you to point additional subdomains at this target (ECS Service), it will create the listener rule for the service on the ALB, and will attempt to create the Alias record in the hosted zone found by `vanity_domain` input.

This adds `additional_targets` input which only does the first part - it adds additional rules to route to this service without attempting to modify hosted zones.

## why

Need to be able to add routes to this service without modifying hosted zone records.


modules/ecs-service/README.md
```diff
@@ -144,6 +144,53 @@ components:
           task_cpu: 256
 ```
 
+#### Other Domains
+
+This component supports alternate service names for your ECS Service through a couple of variables:
+ - `vanity_domain` & `vanity_alias` - This will create a route to the service in the listener rules of the ALB. This will also create a route53 alias record in the hosted zone in this account. The hosted zone is looked up by the `vanity_domain` input.
+ - `additional_targets` - This will create a route to the service in the listener rules of the ALB. This will not create a route53 alias record.
```
milldr, line 151:
```suggestion
 - `additional_targets` - This will create a route to the service in the listener rules of the ALB. This will not create a Route 53 alias record.
```

modules/ecs-service/README.md
```diff
@@ -144,6 +144,53 @@ components:
           task_cpu: 256
 ```
 
+#### Other Domains
+
+This component supports alternate service names for your ECS Service through a couple of variables:
+ - `vanity_domain` & `vanity_alias` - This will create a route to the service in the listener rules of the ALB. This will also create a route53 alias record in the hosted zone in this account. The hosted zone is looked up by the `vanity_domain` input.
```
milldr, line 150:
```suggestion
 - `vanity_domain` & `vanity_alias` - This will create a route to the service in the listener rules of the ALB. This will also create a Route 53 alias record in the hosted zone in this account. The hosted zone is looked up by the `vanity_domain` input.
```

modules/ecs-service/variables.tf
```diff
@@ -196,6 +196,12 @@ variable "vanity_alias" {
   default     = []
 }
 
+variable "additional_targets" {
+  type        = list(string)
+  description = "Additional target routes to add to the ALB that point to this service. The only difference between this and `vanity_alias` is vanity alias will create an alias record in route53 in the hosted zone in this account as well. Additional Targets only adds the listener route to this service's target group."
```
milldr, line 201:
```suggestion
  description = "Additional target routes to add to the ALB that point to this service. The only difference between this and `var.vanity_alias` is `var.vanity_alias` will create an alias record in Route 53 in the hosted zone in this account as well. `var.additional_targets` only adds the listener route to this service's target group."
```

milldr (approved):

PR #956, opened 2024-01-18
Lambda Component Support CICD

Benbentwo:
## what

- Adds changes from our main_overrides.tf to simplify the ability to deploy a lambda through CICD using SSM Parameter store and an s3 bucket

## why

Simplify deployment process


dudymas (approved):
lgtm. Thanks!

PR #963, opened 2024-01-30
`ecs-service` volume passthroughs

Benbentwo:
## what

 - adds passthrough vars for volume support 
 

## why

 - Allows task definitions to declare volumes

## references

 - https://github.com/cloudposse/terraform-aws-ecs-alb-service-task/blob/main/variables.tf#L295-L356


korenyoni (approved):
LGTM.

PR #965, opened 2024-01-30
feat: Multi Org Support for ArgoCD Deployment Repos

milldr:
## what
- Added a new component to create a single GitHub webhook in a single org/repo, `github-webhook`
- Allow the `eks/argocd` component to _not_ create the github webhook
- Handle converting `argocd-repo` component name used in `eks/argocd` to a valid name for helm
- Use the `deepmerge` module for merging maps for notification settings in the `eks/argocd` component
- Allow dynamic list of notifier annotations for `argocd-repo`. Default remains the same

## why
- When deploying `argocd-repo` into different GitHub Organizations, we need to use a different Terraform GitHub provider for each organization. Since Terraform does not support dynamic providers, we have to use multiple components for each provider that we need.
- We can create the GitHub webhook secret value with the `eks/argocd` component (where it's needed for the chart) and store it in SSM. Then we can use this new component `github-webhook` to create the webhook in each repo.
- The webhook secret will be the same for each repo connected to the same argocd deployment
- We need to use `deepmerge` in order to add alternate notifier webhooks without deleting the default options
- We want to be able to customize the list of notifiers for a given `argocd-repo` deployment

## references
- DEV-1567
- Internal PR: https://github.com/cloudposse/infra-live/pull/344


modules/argocd-repo/applicationset.tf
```diff
@@ -1,3 +1,8 @@
+locals {
+  github_default_notifications_enabled = local.enabled && var.github_default_notifications_enabled
```
goruha, line 2:
```suggestion
  github_notifications_enabled = local.enabled && var.github_default_notifications_enabled
```

modules/eks/argocd/github_webhook.tf
```diff
@@ -0,0 +1,74 @@
+# The GitHub webhook can be created with this component, with another component (such as github-webhook), or manually.
+# However, we need to define the value for the webhook secret now with the ArgoCD chart deployment. Store it in SSM for reference.
+#
+# We need to create the webhook with a separate component if we're deploying argocd-repos for multiple GitHub Organizations
+locals {
+  github_webhook_enabled = local.enabled && var.github_webhook_enabled
+  create_github_webhook  = local.github_webhook_enabled && var.create_github_webhook
```
goruha, line 7: What is the difference?

modules/eks/argocd/main.tf
```diff
@@ -1,17 +1,18 @@
...
-  saml_enabled           = local.enabled && var.saml_enabled
+  enabled = module.this.enabled
+
+  kubernetes_namespace = var.kubernetes_namespace
+  oidc_enabled         = local.enabled && var.oidc_enabled
+  oidc_enabled_count   = local.oidc_enabled ? 1 : 0
+  saml_enabled         = local.enabled && var.saml_enabled
   argocd_repositories = local.enabled ? {
-    for k, v in var.argocd_repositories : module.argocd_repo[k].outputs.repository => {
+    for k, v in var.argocd_repositories : replace(k, "/", "-") => {
```
goruha, line 9: Can we leave a reference with `remote-state`?

modules/github-webhook/main.tf
```diff
@@ -0,0 +1,33 @@
...
+  enabled = module.this.enabled
+
+  fetch_github_webhook = local.enabled && !(length(var.webhook_github_secret) > 0)
+
+  # If fetching webhook, get the value from SSM
+  # Else, get the value given by var.webhook_github_secret
+  webhook_github_secret = local.fetch_github_webhook ? try(data.aws_ssm_parameter.webhook[0].value, null) : var.webhook_github_secret
+}
+
+data "aws_ssm_parameter" "webhook" {
```
goruha, line 11: Use `remote-state` pattern to pass the webhook value

modules/eks/argocd/main.tf
```diff
@@ -1,17 +1,18 @@
...
       clone_url         = module.argocd_repo[k].outputs.repository_ssh_clone_url
       github_deploy_key = data.aws_ssm_parameter.github_deploy_key[k].value
+      repository        = module.argocd_repo[k].outputs.repository
     }
   } : {}
-  webhook_github_secret = try(random_password.webhook["github"].result, null)
```
goruha, line 14: How do you pass webhook secret to argocd helm values?

modules/eks/argocd/github_webhook.tf
milldr: `github_webhook_enabled` is whether or not to create a secret value and pass it to the `argo-cd` chart

`create_github_webhook` is whether or not to create a GitHub Webhook for the argocd repos. If you're creating webhooks for multiple organizations and need different github providers, you have to use the `github-webhook` component instead. But you still want to have the same value used for the `argo-cd` chart

> 1. When deploying argocd-repo into different GitHub Organizations, we need to use a different Terraform GitHub provider for each organization. Since Terraform does not support dynamic providers, we have to use multiple components for each provider that we need.
> 2. We can create the GitHub webhook secret value with the eks/argocd component (where it's needed for the chart) and store it in SSM. Then we can use this new component github-webhook to create the webhook in each repo.
> 3. The webhook secret will be the same for each repo connected to the same argocd deployment

https://github.com/cloudposse/terraform-aws-components/blob/46b22591d004c9961daf843abaf49b5ea8e45cdd/modules/eks/argocd/github_webhook.tf#L18-L28

modules/github-webhook/main.tf
milldr: We shouldnt store secrets in terraform output

milldr: I guess we can use `sensitive`

modules/eks/argocd/main.tf
milldr: answered in slack thread. copy paste:


this is still using remote state with argocd-repo
```
  argocd_repositories = local.enabled ? {
    for k, v in var.argocd_repositories : replace(k, "/", "-") => {
      clone_url         = module.argocd_repo[k].outputs.repository_ssh_clone_url
      github_deploy_key = data.aws_ssm_parameter.github_deploy_key[k].value
      repository        = module.argocd_repo[k].outputs.repository
    }
  } : {}
```
the difference is that the key used for the local.argocd_repositories map is the name of the remote component (with a character replacement) instead of module.argocd_repo[k].outputs.repository. Now that value is given to the map as a value for repository instead

modules/github-webhook/main.tf
milldr: added sensitive output and updated `github-webhook` to use either ssm or remote-state. remote-state by default

goruha (approved):

PR #966, opened 2024-01-30
fix-lambda-function-name-null

somansawani:
Solution is to  address a bug in the Terraform lambda component where null values for var.function_name break the formatting of local.cicd_s3_key_format. The bug causes issues when no function name is provided, which should default to using module.label.id instead.

https://github.com/sm-core/infrastructure/pull/1115


milldr (approved):
looks good! thanks for submitting

milldr:
ah this was already merged with #970 last week

https://github.com/cloudposse/terraform-aws-components/blob/main/modules/lambda/main.tf#L7-L9

Issue #962, opened 2024-01-29
[Lambda] null values for var.function_name break the local.cicd_s3_key_format formatting


johncblandii:
### Describe the Bug

`null` values for `var.function_name` break the `local.cicd_s3_key_format` formatting

### Expected Behavior

It uses the `module.label.id` if no function name was provided.

Move the `module.lambda` line for `function_name = coalesce(var.function_name, module.label.id)` to a local and reuse that instead for `cicd_s3_key_format = var.cicd_s3_key_format != null ? var.cicd_s3_key_format : "stage/${module.this.stage}/lambda/${var.function_name}/%s"`

### Steps to Reproduce

Set a `null` `function_name` (or no value) and run a plan.

PR #970, opened 2024-01-31
`lambda` Remove formatting of bucket name

Benbentwo:
## what

 - Removes formatting of the bucket name
 - Uses local for function name which uses `module.id` if not passed in directly

## why

Bucket name variable already uses an essentially hardcoded bucket name, by trying to format it we force the bucket to be in the same account. This isn't part of our recommended arch where the bucket lives in the artifacts account with roles to put objects into the bucket.

## references
 - closes: #962


modules/lambda/main.tf
```diff
@@ -1,9 +1,10 @@
 locals {
-  enabled             = module.this.enabled
-  iam_policy_enabled  = local.enabled && (try(length(var.iam_policy), 0) > 0 || var.policy_json != null)
-  s3_bucket_full_name = var.s3_bucket_name != null ? format("%s-%s-%s-%s-%s", module.this.namespace, module.this.tenant, module.this.environment, module.this.stage, var.s3_bucket_name) : null
```
milldr, line 4: this removes support for dynamically selecting a bucket name. For example

```
components:
  terraform:
     s3-bucket:
        vars:
          name: foo # --> ends up with "%s-%s-%s-%s-%s" name

     lambda:
        vars:
           s3_bucket_name: foo # --> find the "%s-%s-%s-%s-%s" name
```

previously we could dynamically build the rest of the bucket name for any stage

milldr: I dont know if anyone actually uses it that way. Alternatively, we could have a different variable such as `var. s3_bucket_full_name` if you wanted to specify the literal name

Benbentwo: I've re-added bucket_name calculation. I didn't think that the `s3_bucket_name` with calculation was all that helpful. admittedly I forgot about the case where you might want to declare the bucket name once because you actually do have a bucket in each plat account. then we want formatting so you don't need an override per account

milldr (approved):

PR #991, opened 2024-02-25
Use native cloudfront 404 page

goruha:
## what

*  Use native Cloudfront 404 page configuration

## why

* Lambda Edge to generate 404-page cost too much

## references

* https://cloudposse.atlassian.net/browse/DEV-1707


osterman:
We need to discuss this before merging.

osterman (approved):

Issue #977, opened 2024-02-07
aws-backups: variables unused / deprecated


hans-d:
### Describe the Bug

With the upgrade to aws-backup v1 various variables became obsolote, buta re not marked as such

### Expected Behavior

variables functioning,and/or marked deprecated, or fully removed

### Steps to Reproduce

see code

### Additional Context

https://github.com/cloudposse/terraform-aws-components/blob/633f81082fb13cdf5febebdf5aee77505d585e04/modules/aws-backup/variables.tf#L18

hans-d:
Example in the readme must also be updated

oycyc:
Saw this issue after trying to use this component, made a PR for this: https://github.com/cloudposse/terraform-aws-components/pull/992

oycyc:
https://github.com/cloudposse/terraform-aws-components/pull/992 is merged

hans-d:
thanks @oycyc

PR #992, opened 2024-02-26
remove unused vars in `aws-backup` component to reflect the module upgrade

oycyc:
## what

After the [AWS Backup module https://github.com/cloudposse/terraform-aws-backup/](https://github.com/cloudposse/terraform-aws-backup/) 1.0 release, there's a couple of unused variables and unused logic, and some information that should be updated in the README. 

- The variables `schedule`, `start_window`, `completion_window`, `cold_storage_after`, `delet_after`, and `destination_vault_arn`, etc. were removed as of the module's v0.14 and placed in deprecation (see PR https://github.com/cloudposse/terraform-aws-backup/pull/39/files#diff-0c2612bc0fbe3bc116da52a6aece40330dc93c8cbc640dc96427bf940f08f7c3) then finally removed entirely when the v1.0 release (see this PR by @Benbentwo : https://github.com/cloudposse/terraform-aws-backup/pull/63)
  - They were moved to the `rules` object to support multiple plans, so getting rid of them in this component since they're not used anywhere anymore.
  - Because the `destination_vault` and related variables aren't used on the top level, the `remote-state.tf` file and the `locals` block that dynamically determines if the copy action is enabled is not used anymore either here.

Also updated the README to reflect these changes along with the new method for cross region copying.

## why

Unused variables taking up space & tflint complaining 😅

Also an existing issue for this: https://github.com/cloudposse/terraform-aws-components/issues/977

## references

See links above for each point.


Gowiem:
@hans-d we don't have terratest for components unfortunately 😞

Gowiem (approved):
LGTM! We'll need a second review from someone on the @cloudposse/admins team though.

PR #998, opened 2024-03-08
feat: `prettier` in READMEs

milldr:
## what
- Added `prettier` to format all Markdown files
- Added ignore wrapper for `terraform-docs` formatted tables
- Added new lines after all admonitions as such:

```bash
#!/bin/bash

replace_strings=(
"<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->/<!-- prettier-ignore-start -->\n<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->"
"<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->/<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->\n<!-- prettier-ignore-end -->"
":::info/:::info\n"
":::important/:::important\n"
":::warning/:::warning\n"
":::caution/:::caution\n"
)

for rs in "${replace_strings[@]}"
do
  echo ${rs}
  LC_ALL=C find terraform-aws-components/modules -name README.md -type f -exec sed -r -i '' "s/${rs}/g" {} +
done
```

- Ignore primary README and all deprecated content

## why
- Prettier is an opinionated code formatter. In particular, for markdown
- Easily reversible: delete it later if we don't like it
- Terraform Docs and Prettier would compete to format the same content. We want to ignore the content that is created by Terraform Docs

## ref
- https://prettier.io/docs/en/
- Internal discussion: https://github.com/orgs/cloudposse-corp/discussions/233


dudymas (approved):
I honestly scanned a lot of this... but nothing seems out of place. Thanks

PR #999, opened 2024-03-11
Fix: `ecs-service` perma-drift for `key_id`

milldr:
## what
- Only set `key_id` for `aws_ssm_parameter` when the type is "SecureString"

## why
- With other types Terraform will pass but will constantly suggest adding the `key_id`

## references
- https://github.com/hashicorp/terraform-provider-aws/issues/21095


korenyoni (approved):
lgtm

PR #1003, opened 2024-03-15
chore: update module in datadog-monitor

dudymas:
## what

- update monitor module in datadog-monitor component
- update readme for datadog-monitor

## why

- changed to support importing monitors


modules/datadog-monitor/README.md
```diff
@@ -89,29 +94,29 @@ elb-lb-httpcode-5xx-notify:
     Check LB
   escalation_message: ""
   tags: {}
+  options:
+    priority: 3
```
Benbentwo, line 98:
```suggestion
```

Looks like `priority` is not a part of options but it's own variable

dudymas: yep! I missed that. Thanks!

PR #1006, opened 2024-03-25
feat: add org compliance packs to aws-config

dudymas:
## what

- new module in `aws-config` for org scoped conformance packs

## why

- allows conformance packs to automatically be applied to accounts in the org


modules/aws-config/main.tf
```diff
@@ -40,11 +40,32 @@ module "utils" {
   context = module.this.context
 }
 
+locals {
+  account_enabled = local.enabled && var.scope == "account"
+  org_enabled     = local.enabled && var.scope == "organization"
+}
+
```
mcalhoun, line 47: Doesn't this approach prevent us from mixing account conformance packs from organizational ones? What if we only want certain conformance packs in PHI accounts, for example, but want others for the whole org?

I think it would be better to add a flag `var.conformance_packs` to indicate if it's an org conformance pack and default to account if not specified.

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Usage
+
+First, make sure your root `account` allows the service access principal `config-multiaccountsetup.amazonaws.com` to
+update child organizations. You can see the docs on the account module here:
+[aws_service_Access_principals](https://docs.cloudposse.com/components/library/aws/account/#input_aws_service_access_principals)
+
+Then set the `scope` of the parent `aws-config` to be `organization`.
+
+After that, any conformance packs you define in the `conformance_packs` variable will be deployed to all child accounts
+of the organization.
```
mcalhoun, line 21:
```suggestion
After that, any conformance packs you define in the `conformance_packs` variable will be deployed to all child accounts of the organization.
```

mcalhoun (changes requested):
I suggested a different approach

modules/aws-config/main.tf
dudymas: almost done with this change. Testing it locally

dudymas: I found that setting a `default_scope` should be supported. @mcalhoun, I'll resolve this for now. I investigated the pack yaml's like you mentioned, and it looks like they are able to handle either scenario. I don't see any regressions with my changes. If you'd like to test this internally and make sure, I'm happy to help and to also show my own work (I've internally tested against a live aws-config component)

modules/aws-config/modules/org-conformance-pack/README.md
dudymas: refactor has removed this statement entirely

mcalhoun (approved):
LGTM!

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
+# AWS Config Conformance Pack
+
+This module deploys a
+[Conformance Pack](https://docs.aws.amazon.com/config/latest/developerguide/conformance-packs.html). A conformance pack
+is a collection of AWS Config rules and remediation actions that can be easily deployed as a single entity in an account
+and a Region or across an organization in AWS Organizations. Conformance packs are created by authoring a YAML template
+that contains the list of AWS Config managed or custom rules and remediation actions.
+
+The Conformance Pack cannot be deployed until AWS Config is deployed, which can be deployed using the
+[root module](../../) of this repository.
```
osterman, line 10: of this repository? The `../../` makes it more clear, but seems we should call it the `aws-confg` component.

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
+# AWS Config Conformance Pack
+
+This module deploys a
+[Conformance Pack](https://docs.aws.amazon.com/config/latest/developerguide/conformance-packs.html). A conformance pack
```
osterman, line 4: Awkward line breaks

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+that contains the list of AWS Config managed or custom rules and remediation actions.
+
+The Conformance Pack cannot be deployed until AWS Config is deployed, which can be deployed using the
+[root module](../../) of this repository.
+
+## Usage
+
+First, make sure your root `account` allows the service access principal `config-multiaccountsetup.amazonaws.com` to
+update child organizations. You can see the docs on the account module here:
+[aws_service_Access_principals](https://docs.cloudposse.com/components/library/aws/account/#input_aws_service_access_principals)
```
osterman, line 16:
```suggestion
[aws_service_access_principals](https://docs.cloudposse.com/components/library/aws/account/#input_aws_service_access_principals)
```

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Then you have two options:
+
+- set the `default_scope` of the parent `aws-config` component to be `organization` (can be overridden by the `scope` of
```
osterman, line 20:
```suggestion
- Set the `default_scope` of the parent `aws-config` component to be `organization` (can be overridden by the `scope` of
```

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+  each `conformance_packs` item)
+- set the `scope` of the `conformance_packs` item to be `organization`
+
+An example yaml stack config for atmos is as follows (both options are shown for demonstration purposes. In practice you
```
osterman, line 24:
```suggestion
An example YAML stack config for Atmos follows. Note, that both options are shown for demonstration purposes. In practice you
```

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+should only have one `aws-config` per account):
```
osterman, line 25:
```suggestion
should only have one `aws-config` per account:
```

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```yaml
+components:
+  terraform:
+    account:
+      vars:
+        aws_service_access_principals:
+          - config-multiaccountsetup.amazonaws.com
+
+    aws-config/example/1:
```
osterman, line 35:
```suggestion
    aws-config/cis:
```

modules/aws-config/modules/org-conformance-pack/README.md
```diff
@@ -0,0 +1,48 @@
...
+      vars:
+        conformance_packs:
+          - name: Operational-Best-Practices-for-CIS-AWS-v1.4-Level1
+            conformance_pack: https://raw.githubusercontent.com/awslabs/aws-config-rules/master/aws-config-conformance-packs/Operational-Best-Practices-for-CIS-AWS-v1.4-Level1.yaml
+            scope: organization
+
+    aws-config/example/2:
```
osterman, line 42:
```suggestion
    aws-config/cis/level-2:
```

modules/aws-config/README.md
```diff
@@ -20,6 +20,22 @@ Some of the key features of AWS Config include:
...
+    through rules in the AWS Config interface to see which rules are not being evaluated.
+  - If you end up still needing more than 1000 rules, one recommendation is to only run packs on a schedule with a
+    lambda that removes the pack after results are collected. If you had different schedule for each day of the week,
+    that would mean 7000 rules over the week. The aggregators would not be able to handle this, so you would need to
+    make sure to store them somewhere else (i.e. S3) so the findings are not lost.
+  - See the
+    [Audit Manager docs](https://aws.amazon.com/blogs/mt/integrate-across-the-three-lines-model-part-2-transform-aws-config-conformance-packs-into-aws-audit-manager-assessments/)
+    if you think you would like to convert conformance packs to custom Audit Manager assessments.
+- The maximum number of AWS Config conformance packs that can be created in a single account is 50.
+
```
osterman, line 38: Use an admonition

modules/aws-config/README.md
```diff
@@ -20,6 +20,22 @@ Some of the key features of AWS Config include:
 - Notifications and alerts: AWS Config can send notifications and alerts when changes are made to your AWS resources
   that could impact their compliance or security posture.
 
+You'll also want to be aware of some limitations with AWS Config:
```
osterman, line 23:
```suggestion
## AWS Config Limitations

You'll also want to be aware of some limitations with AWS Config:
```

modules/aws-config/modules/org-conformance-pack/README.md
dudymas: this isn't a very actionable comment. Would you like everything to be on an extra long line? Or would you rather the link be separated from the text?

dudymas: @osterman let me know if there's some hard and fast rule to handling this. To be clear, it doesn't seem to affect the [rendered result](https://github.com/cloudposse/terraform-aws-components/blob/b5092ac4851a582f3c56232073370526e18bb476/modules/aws-config/modules/org-conformance-pack/README.md)

osterman: Fair point, I don't have a rule or a way to codify it. 

For example, this would also render the same way:

```
is a collection 
of AWS Config rules and remediation 
actions 
that can be easily deployed as 
a single entity in an account
and a Region or across an organization in AWS Organizations. 
Conformance packs are created by authoring a YAML template
that contains 
the list of AWS 
Config managed or custom 
rules and remediation actions.
```

But that would be a bit awkward, IMO.

My point is, if you were not width constrained, it's probably not where the line break would naturally be placed. Use natural line breaks, not 80-characters or other arbitrary constraints. You're absolutely right, it's a minor point.

osterman (approved):

modules/aws-config/modules/org-conformance-pack/README.md
dudymas: So, I tried reflowing this and prettier snapped it back to this version. For now, I think I'll just leave it, and instead we'll need to look at the prettier config. maybe @milldr knows why this is happening?

PR #1019, opened 2024-04-16
feat(aurora-pg): allow passing extra SGs to allow

Gowiem:
## what

* Adds passing additional security group names to the `aurora-postgres` component to support additional inbound rules into the resulting aurora PG cluster. 

## why

* This is needed when you have external systems that you want to have access to the DB cluster that are not accounted for by EKS and you don't want to be overly permissive with allowing a the full CIDR Block. In our case, we're using AWS DMS and need that to be able to communicate with one of our DB clusters. 

## references

* N/A


joe-niland (approved):
LGTM!

milldr (approved):

PR #1024, opened 2024-04-19
feat: `ec2-instance` Component

milldr:
## what
- Create a new component for `ec2-instance`

## why
- This component implements the most basic usage of an EC2 instance

## references
- Customer engagement


modules/ec2-instance/templates/userdata.sh.tmpl
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo "Hello user data"
```
Benbentwo, line 3: Shouldn't this include some variable and better support templating?

e.g. 
```bash
#!/bin/bash

echo "Begin userdata.sh"
${var.userdata}
```

Benbentwo (approved):
Doesn't look like the userdata.sh template is super helpful now. but this component can still be used for spinning up a singular instance as is.

modules/ec2-instance/templates/userdata.sh.tmpl
milldr: a lot of this was taken from the `zscaler` component, so I had stripped what wasnt necessary for a basic ec2 instance. But that's a good idea. Will add that in

Benbentwo (approved):

PR #1025, opened 2024-04-28
feat: add default_tags

nitrocode:
## what

- [x] add default_tags
- [x] self reviewed

## why

- Consistent tagging without needing to try
- Take advantage of the fixed `default_tags` in aws v5 provider

## references

- https://www.hashicorp.com/blog/terraform-aws-provider-5-0-adds-updates-to-default-tags
- https://sweetops.slack.com/archives/C031919U8A0/p1714328473235309


osterman:
@nitrocode this looks interesting, but it's too risky to rollout en masse. We're actively working on rolling out new testing infrastructure for components and will be moving them individually to a new GitHub org.

Additionally, by enabling this, it makes it impossible within a component to _not_ assign certain tags to resources.


There is a workaround, if you would still like this behavior.

- Use [atmos to generate provider tags](https://atmos.tools/core-concepts/components/terraform/providers/#provider-configuration-and-overrides-in-atmos-manifests)

nitrocode:
Interesting. I wonder why anyone would not want to tag a resource.

I'll check that out. Thank you

PR #1037, opened 2024-05-15
`sqs-queue` better support dead-letter queues

Benbentwo:
## what
* Updates component `sqs-queue` to better support dead-letter queues

## why

* This is an awkward component currently, it is really just a pass through to an sqs resource. This should probably be broken down into a module and then wrapped with a component. In the meantime, this will allow easier access to other sqs-queues' URL for the redrive policy which is used for deadletter queues

## references
 * https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue_redrive_policy


modules/sqs-queue/variables.tf
```diff
@@ -41,10 +41,28 @@ variable "policy" {
...
+  description = "The name of the component that will be looked up for the ARN and be used as the dead letter queue."
+  default     = null
+}
+
+variable "dead_letter_max_receive_count" {
+  type        = number
+  description = "The number of times a message can be unsuccessfully dequeued before being moved to the dead-letter queue."
+  default     = 5
+}
+
```
korenyoni, line 65: nit: other variables capitalize SQS and Dead Letter Queue

modules/sqs-queue/remote-state.tf
```diff
@@ -0,0 +1,11 @@
+
+module "dead_letter_sqs_remote_state" {
+  count = var.dead_letter_sqs_component_name != null ? 1 : 0
```
korenyoni, line 3: Maybe to be safe with `""` we should do

```suggestion
  count = try(length(var.dead_letter_sqs_component_name), 0) > 0 ? 1 : 0
```

modules/sqs-queue/main.tf
```diff
@@ -1,7 +1,8 @@
 locals {
-  enabled            = module.this.enabled
-  aws_account_number = one(data.aws_caller_identity.current[*].account_id)
-  policy_enabled     = local.enabled && length(var.iam_policy) > 0
+  enabled                = module.this.enabled
+  aws_account_number     = one(data.aws_caller_identity.current[*].account_id)
+  policy_enabled         = local.enabled && length(var.iam_policy) > 0
+  redrive_policy_enabled = local.enabled && (var.dead_letter_sqs_component_name != null || var.dead_letter_sqs_url != null)
```
korenyoni, line 5: See comment on `try(length(),0)` below

modules/sqs-queue/main.tf
```diff
@@ -69,3 +70,13 @@ resource "aws_sqs_queue_policy" "sqs_queue_policy" {
   queue_url = module.sqs_queue.url
   policy    = one(module.queue_policy[*].json)
 }
+
+resource "aws_sqs_queue_redrive_policy" "dead_letter_queue" {
+  count = local.redrive_policy_enabled != null ? 1 : 0
+
+  queue_url = module.sqs_queue.url
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = var.dead_letter_sqs_url != null ? var.dead_letter_sqs_url : one(module.dead_letter_sqs_remote_state[*].outputs.arn)
```
dudymas, line 79: Just a sanity check here... This looks like you're assigning a URL to an ARN. Don't you need to look up the sqs queue with [a data query](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/sqs_queue) if you want the ARN? Otherwise, I'm not sure this would work.

modules/sqs-queue/remote-state.tf
Benbentwo: I think this is safe enough, the variable is defaulted to null, so someone would have to be explicitly trying to set the component name to `""`.

modules/sqs-queue/main.tf
Benbentwo: Misnamed the variable, that should've been an arn.

korenyoni (approved):

PR #1042, opened 2024-05-22
SQS Component Refactor

Benbentwo:
## what
* Refactor SQS Component to utilize `"terraform-aws-modules/sqs/aws"` module
* Add IAM Policy Support

## why
* Better support for dlq within the same component
* Policy enhancements

## references
* N/A


modules/sqs-queue/CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
+## Upgrading to `v1.TODO.0`
```
milldr, line 1: I believe we decided to just call this the PR # since we cant determine the version until it's released of course

milldr (approved):

PR #1044, opened 2024-05-22
feat: `elasticache-redis` Optionally Disable VPC Ingress Rule

milldr:
## what
- Add new variable 'allow_ingress_from_this_vpc' for `elasticache-redis`
- ~~Changed default behavior to not allow Ingress from the VPC in this account~~ removed

## why
- Previously the default behavior allowed the full VPC CIDR in this account without an option to disable it. We should not only be able to disable it if we choose, ~~but the default behavior should be _less_ permissive.~~

## references
- Customer ask


modules/elasticache-redis/variables.tf
```diff
@@ -75,6 +75,12 @@ variable "redis_clusters" {
   description = "Redis cluster configuration"
 }
 
+variable "allow_ingress_from_this_vpc" {
+  type        = bool
+  default     = false
```
Nuru, line 80:
```suggestion
  default     = true
```

We should not change default behavior except for:
- Major upgrades, with big warnings about breaking changes
- Previous default behavior no longer works or is being exploited by malicious users.

Nuru (changes requested):

modules/elasticache-redis/variables.tf
milldr: fair enough. Updated

dudymas (approved):
lgtm

Nuru (approved):

PR #1058, opened 2024-06-04
Update bats workflow

goruha:
## what
* Replace `pull_request_target` by `pull_request`
* Use `terraform 1` for bats

## why
* Improve security 
* test-harness support terraform and opentofu, use terraform for bats

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -1,7 +1,7 @@
 name: bats
 
 on:
-  pull_request_target:
+  pull_request:
     types: [labeled, opened, synchronize, unlabeled]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/bats.yml
```diff
@@ -32,6 +28,7 @@ jobs:
         HEAD_REF: ${{ github.head_ref }}
       run: |
         # when running in test-harness, need to mark the directory safe for git operations
+        echo "/usr/share/terraform/1/bin/terraform" >> "$GITHUB_PATH"
```
Nuru, line 31: This should not be necessary, because we should continue to be (and currently are) installing a link to the default version, even if it is tofu, as `/usr/bin/terraform`. 

```suggestion
```

Nuru (dismissed):

.github/workflows/bats.yml
goruha: Unfortunately, we can not do as usual.  Terratest fallback to opentofu if terraform binary not found. 
Otherwise, we must update all terraform modules `go` code to support explicit binary path.

Nuru: This workflow is not running Terratest, it is running `bats`, which uses whatever `terraform` is installed.

osterman (approved):

PR #1063, opened 2024-06-09
feat(aurora-postgres): backup configs

oycyc:
## what

The CloudPosse RDS cluster module https://github.com/cloudposse/terraform-aws-rds-cluster has these backup configurations that would be helpful to have!

## why

1. Allows me to set backup configurations
2. Without these inputs, the cluster is forced to use the default configs from the module
3. Allows me to set these variable values to `null` or empty when I don't want any backing up happening
4. Without these inputs to set it as empty values, there's a conflict when using RDS clusters with AWS Backup plans because the RDS resource maintains its backup values, while AWS Backup plans wants other values.

## references
- CloudPosse's RDS module
- 
image
- AWS Backup conflict with default backup values unless explictly set https://github.com/hashicorp/terraform-provider-aws/issues/33465


Benbentwo (approved):
Thanks for handling this!

oycyc:
Wanted to a quick check up on this so it doesn't get stale! Seems like I can't merge it myself

Benbentwo:
Thanks for re-commenting, got lost in slack and other GH issues. Re-ran everything and it looks good!

oycyc:
Thanks!

PR #1074, opened 2024-07-03
Rebuild readmes

osterman:
## what

- Implement a `terraform-docs` template

## why

- Fix formatting to be compatible with MDX


goruha:
We will have that as part of default workflows in new org

PR #1080, opened 2024-07-22
feat: add additional variables and outputs for `spa-s3-cloudfront`

korenyoni:
## what

* add `origin_bucket` variable
* add `s3_origins` variable
* add `cloudfront_distribution_identity_arn` output

## why

* Add variables and outputs present in the [cloudposse/terraform-aws-cloudfront-s3-cdn](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/blob/main/variables.tf) module but missing in this component

## references

N/A


dudymas (approved):
Please note, this doesn't allow [custom_origin_config](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution.html#custom_origin_config)... As long as you're confident that only s3 needs to be supported, I'd say go right ahead.

korenyoni:
> Please note, this doesn't allow [custom_origin_config](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution.html#custom_origin_config)... As long as you're confident that only s3 needs to be supported, I'd say go right ahead.

So actually, we already have `var.custom_origins` handling that 

https://github.com/cloudposse/terraform-aws-components/blob/671f4db5512133515e71f4fc739291a763142f4c/modules/spa-s3-cloudfront/main.tf#L131

PR #1088, opened 2024-08-01
Added additional polices for vpn and kms - required by planner

goruha:
## what
* Added VPN export reader policy
* Added KMS decrypt policy

## why
Policies required for gitops dynamic terraform roles planner

## references
* https://github.com/cloudposse/infra-live/actions/runs/10181554256/job/28167900030
* https://github.com/cloudposse/infra-live/actions/runs/10181554256/job/28164592870

### DIFF
--- a/modules/aws-team-roles/policy-kms-planner.tf.tf
+++ b/modules/aws-team-roles/policy-kms-planner.tf.tf
@@ -1,0 +1,36 @@
+locals {
+  kms_planner_enabled = contains(local.configured_policies, "kms_planner")
+}
+
+data "aws_iam_policy_document" "kms_planner_access" {
+  count = local.kms_planner_enabled ? 1 : 0
+
+  statement {
+    sid    = "AllowKMSDecrypt"
+    effect = "Allow"
+    actions = [
+      "kms:Decrypt",
+    ]
+    resources = [
+      "*"
+    ]
+  }
+
+}
+
+data "aws_iam_policy_document" "kms_planner_access_aggregated" {
+  count = local.kms_planner_enabled ? 1 : 0
+
+  source_policy_documents = [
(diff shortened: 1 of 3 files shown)


modules/aws-team-roles/policy-kms-planner.tf.tf
```diff
@@ -0,0 +1,36 @@
...
+data "aws_iam_policy_document" "kms_planner_access" {
+  count = local.kms_planner_enabled ? 1 : 0
+
+  statement {
+    sid    = "AllowKMSDecrypt"
+    effect = "Allow"
+    actions = [
+      "kms:Decrypt",
+    ]
+    resources = [
```
Nuru, line 14: Restrict decryption to SSM parameters
```suggestion
    
    # Only allow decryption of SSM parameters.
    # To further restrict to specific parameters, add conditions on the value of 
    #   kms:EncryptionContext:PARAMETER_ARN
    # See https://docs.aws.amazon.com/kms/latest/developerguide/services-parameter-store.html#parameter-store-encryption-context
    condition {
      test     = "Null"
      variable = "kms:EncryptionContext:PARAMETER_ARN"
      values   = ["false"]
     }
     
     resources = [
```

Nuru (changes requested):
To start with, we can limit KMS key access to decrypting SSM parameters. Later we can add a whitelist or blacklist of parameters to decrypt, and we can similarly enable other decryption operations.

Still, I think we should have a plan for somehow marking components as too sensitive to plan and moving forward with the rest of GitOps as if those plans were successful without actually attempting them.

Nuru (approved):

Nuru (approved):

milldr (approved):

PR #1136, opened 2024-10-03
feat(kms): updates with embedded policy creation - part2

Gowiem:
## what

* Updates the KMS module to support embedded policy creation

## why

* This allows for easy wiring in of aws-team-role roles into the KMS policy, so we do something like "Admins in the dev account have access to use this Key"

## references

* This is a continuation of PR #523 which I could not continue as I no longer have access to push to Cloud Posse except as a fork 😢 
  * @nitrocode did a great review with some solid suggestions in #523. I have completed my work from that review as part of this newer PR.


modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
+locals {
+  account_id        = data.aws_caller_identity.current.account_id
+  account_principal = "arn:${data.aws_partition.current.partition}:iam::${local.account_id}:root"
```
nitrocode, line 3: In the specified account, this would make every IAM role and user an administrator of the key.

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
+
+    actions = [
+      "kms:*",
+    ]
+
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = [local.account_principal]
```
nitrocode, line 55: Related to https://github.com/cloudposse/terraform-aws-components/pull/1136/files#r1785624887

You could have an input that would allow overriding this for least privilege.

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
+    sid    = "KeyUsage"
+    effect = "Allow"
+
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
+      "kms:DescribeKey",
+    ]
```
nitrocode, line 69: Sometimes, such as in the case of SSM customer managed KMS keys, you may only want these perms

```
    actions = [
      "kms:Decrypt",
      "kms:GenerateDataKey*",
    ]
```

Perhaps these actions can be overridable too?

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
+locals {
+  account_id        = data.aws_caller_identity.current.account_id
+  account_principal = "arn:${data.aws_partition.current.partition}:iam::${local.account_id}:root"
+
+  principals = sort(distinct(concat(
```
nitrocode, line 5: I'd suggest keeping the original name so it's clearer in the kms policy

```suggestion
  allowed_principal_arns = sort(distinct(concat(
```

or making it even clearer

```suggestion
  kms_usage_principal_arns = sort(distinct(concat(
```

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
+      "kms:DescribeKey",
+    ]
+
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = local.principals
```
nitrocode, line 75: Related comment https://github.com/cloudposse/terraform-aws-components/pull/1136/files#r1785626505

modules/kms/main.tf
nitrocode, line 1: Have you tested this component with `enabled: false` to see if it still plans?

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
+
+data "aws_caller_identity" "current" {}
+
+data "aws_partition" "current" {}
+
+module "allowed_role_map" {
+  source = "../account-map/modules/roles-to-principals"
+
+  privileged = false
+  role_map   = var.allowed_roles
```
nitrocode, line 19: nit: why not `allowed_role_map` for consistency with the module input ?

```suggestion
  role_map   = var.allowed_role_map
```

Other examples in other components

https://github.com/cloudposse/terraform-aws-components/blob/f51b6f31b196f1cfa3befe7f17024b1957bb8eeb/modules/eks/cluster/remote-state.tf#L7-L10

https://github.com/cloudposse/terraform-aws-components/blob/f51b6f31b196f1cfa3befe7f17024b1957bb8eeb/modules/ecr/main.tf#L1-L4

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
   customer_master_key_spec = var.customer_master_key_spec
   multi_region             = var.multi_region
 
+  policy = var.policy != "" ? var.policy : data.aws_iam_policy_document.key_policy.json
+
   context = module.this.context
 }
+
+data "aws_iam_policy_document" "key_policy" {
+
```
nitrocode, line 42: nit

```suggestion
```

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
   key_usage                = var.key_usage
   customer_master_key_spec = var.customer_master_key_spec
   multi_region             = var.multi_region
 
+  policy = var.policy != "" ? var.policy : data.aws_iam_policy_document.key_policy.json
+
   context = module.this.context
 }
+
+data "aws_iam_policy_document" "key_policy" {
```
nitrocode, line 41: nit: this name is redundant

- the `key` portion is redundant because this is the `kms` component
- the `policy` portion is redundant because this is the `data.aws_iam_policy_document`

I realize that the `module.kms_key` is right above it which is in violation of my first bullet nitpick but because it's pre-existing and what we do in most components, it's roughly consistent. Since this is a net-new data source, we can change it to `default` or `this`.

```suggestion
data "aws_iam_policy_document" "default" {
```

modules/kms/main.tf
```diff
@@ -1,15 +1,78 @@
...
+
+  principals = sort(distinct(concat(
+    var.allowed_principal_arns,
+    module.allowed_role_map.principals,
+  )))
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_partition" "current" {}
```
nitrocode, line 13: Related to https://github.com/cloudposse/terraform-aws-components/pull/1136/files#r1785627233

Here is an example of similar data sources turned off if the `enabled` flag is false

https://github.com/cloudposse/terraform-aws-components/blob/f51b6f31b196f1cfa3befe7f17024b1957bb8eeb/modules/aws-inspector2/main.tf#L20-L22

```suggestion
data "aws_caller_identity" "current" {
  count = module.this.enabled ? 1 : 0
}

data "aws_partition" "current" {
  count = module.this.enabled ? 1 : 0
}
```

modules/kms/README.md
nitrocode, line 1: Here's an alternative idea, feel free to ignore it and push through with your PR.

There is an ask from a community member to have canned policies https://github.com/cloudposse/terraform-aws-kms-key/issues/25 similar to the [S3 bucket module](https://github.com/cloudposse/terraform-aws-s3-bucket/blob/a316dfb199c9d830b803e36e8f64f6661ace6b7c/main.tf#L363) and the [ECR module](https://github.com/cloudposse/terraform-aws-ecr/blob/d50edfb3860ca6fdfa6737a38b2fe3115e0a14dc/main.tf#L126). If canned policies were in the module itself, new inputs could be exposed, and you'd need to do less work in the component as more of the logic would exist in the upstream module.

nitrocode:
Thanks for addressing all the changes @Gowiem. Feel free to ignore my comments as I do not have review power here. I got the @ sign notification so thought I'd re-review.

goruha (approved):

goruha (approved):

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

modules/kms/main.tf
goruha: Ignored

goruha:
@Benbentwo, we need to address all PRs before codefreeze. So, I liberate the criteria

PR #1146, opened 2024-10-08
fix: ssosync artifact filepath extraction

joelmccoy:
## what

Fixes the ssosync module to use the absolute filepath when extracting the tar.  The code makes some assumptions that relative path works fine.  But when you get into nesting modules, you can run into errors.

## why

* simple bug fix allows me to call this module directly


joelmccoy:
this actually doesn't seem to be a straight forward fix.  closing out this PR and reevaluating.

PR #1147, opened 2024-10-08
upstream fixes for site-to-site-vpn

Benbentwo:
## what

 - Couple issues remained
   - `null` variable (when set to `nullable=false`)
   - misnamed module
## why

 - Bugfixes

## references

#1144

### DIFF
--- a/modules/site-to-site-vpn/main.tf
+++ b/modules/site-to-site-vpn/main.tf
@@ -15,7 +15,7 @@ locals {
   ) : null
 }
 
-module "site_to_site_vpn" {
+module "vpn_connection" {
   source  = "cloudposse/vpn-connection/aws"
   version = "1.3.1"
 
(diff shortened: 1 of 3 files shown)


modules/site-to-site-vpn/main.tf
```diff
@@ -15,7 +15,7 @@ locals {
   ) : null
 }
 
-module "site_to_site_vpn" {
```
milldr, line 18: this'll cause a destroy and replace for anyone with this component currently deployed. Although this isnt particular common, we should at least include a changelog to call this out and warn anyone vendoring

milldr (changes requested):

modules/site-to-site-vpn/main.tf
Benbentwo: This should actually be a fix, the outputs of this component refer to `module.vpn_connection` - which is this module, or meant to be. So it was not deployable without renaming it.

milldr (approved):

PR #1162, opened 2024-10-16
Allow removal of cluster name from aurora ssm parameter

darrenl33:
## what

- ssm parameter is prefixed with the cluster name, adding an cluster_name_override to not set
- allows prefix to follow naming convention for other parameters by only using ssm_path_prefix


modules/aurora-postgres/ssm.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  ssm_path_prefix = format("/%s/%s", var.ssm_path_prefix, module.cluster.id)
+  ssm_path_prefix = var.ssm_cluster_name_override ? format("/%s", var.ssm_path_prefix) : format("/%s/%s", var.ssm_path_prefix, module.cluster.id)
```
Benbentwo, line 2:
```suggestion
  ssm_path_prefix = var.ssm_cluster_name_override ? format("/%s/%s", var.ssm_path_prefix, var.ssm_cluster_name_override) : format("/%s/%s", var.ssm_path_prefix, module.cluster.id)
```

I might be missing something from this PR but wouldn't it make sense to utilize the override being provided?

darrenl33: make sense, updated

Benbentwo:
Does this work for you? 

> allows prefix to follow naming convention for other parameters by only using ssm_path_prefix

this currently would still have something between the `ssm_path_prefix` and the `/user`.

goruha:
@darrenl33, why do we need this?
Could you pls provide any use-cases for this PR?

goruha (changes requested):
Check the comment

modules/aurora-postgres/ssm.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  ssm_path_prefix = format("/%s/%s", var.ssm_path_prefix, module.cluster.id)
+  ssm_path_prefix = length(var.ssm_cluster_name_override) > 0  ? format("/%s/%s", var.ssm_path_prefix, var.ssm_cluster_name_override) : format("/%s/%s", var.ssm_path_prefix, module.cluster.id)
```
goruha, line 2:
```suggestion
  ssm_path_prefix = format(var.ssm_path_prefix_format, module.cluster.id)
```

Remove variable `ssm_path_prefix`
Add variable `ssm_path_prefix_format` with default value `"/aurora-postgres/%s"`

Than you will be able to override it

goruha (changes requested):

modules/aurora-postgres/ssm.tf
goruha: or to support backcompatibility

```
ssm_path_prefix = format(var.ssm_path_prefix_format, var.ssm_path_prefix, module.cluster.id)
```

Add variable ssm_path_prefix_format with default value "/%s/%s"

darrenl33: this will still append module.cluster.id? so instead of the cluster_id that might look like `namespace+tenant+environment+service_name-main` , we wanted a shorter parameter store name, so if ssm_cluster_name_override, we'll use that instead.

goruha: you can set `ssm_path_prefix_format="/aurora-postgres/short-cluster-name"`
and that would be the same

goruha (approved):

PR #1164, opened 2024-10-17
feat(efs): allow specifying environment for dns-delegated component

marsadle:
## what

Add `dns_gbl_delegated_environment_name` (used throughout components) to specify environment of `dns-delegated` component.

## why

This change is necessary when deploying identical environments in multiple regions and using regional `dns-delegated` components rather than a single global one.


marsadle:
Closing as similar to #1165.

PR #1165, opened 2024-10-17
feat(elasticache-redis): allow specifying environment for dns-delegated component

marsadle:
## what

Add `dns_gbl_delegated_environment_name` (used throughout components) to specify environment of `dns-delegated` component.

## why

This change is necessary when deploying identical environments in multiple regions and using regional `dns-delegated` components rather than a single global one.

### DIFF
--- a/modules/elasticache-redis/remote-state.tf
+++ b/modules/elasticache-redis/remote-state.tf
@@ -23,7 +23,7 @@ module "dns_delegated" {
   version = "1.5.0"
 
   component   = "dns-delegated"
-  environment = "gbl"
+  environment = var.dns_gbl_delegated_environment_name
 
   context = module.this.context
 }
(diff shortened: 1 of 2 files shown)


Nuru:
@marsadle Why are you using regional dns-delegated components? DNS is not regional in any way, and our architecture is set up for you to use a single dns-delegated component per account.

marsadle:
Your `dns-delegated` component actually describes, that it can be deployed regionally.
Not all components are set up to do so, e.g. `aurora-postgres` which has no way to point to a regional endpoint. We need `postgres-writer.euw1.our-infrastructure.cloud` and `postgres-writer.use1.our-infrastructure.cloud` which are entirely separate environments. The component only provisions `postgres-writer.our-infrastructure.cloud` when provisioned with a global `dns-delegated` zone.

This is a change that is non-intrusive and has been made across components (e.g. [aurora-postgres](https://github.com/cloudposse/terraform-aws-components/blob/main/modules/aurora-postgres/remote-state.tf#L44)), which is much easier to do than to alter components into setting custom DNS names.

PR #1170, opened 2024-10-29
feat(elasticache-redis): add engine input for valkey support

nitrocode:
## what

- add engine input for valkey support

## why

- Valkey is far cheaper than redis

<details><summary>Notes</summary>

- Design options
    1. Add a new key to `local.cluster_attributes` and add a variable to the component, default it to `redis`, and pass as an argument to the module as-is
        - We could shy away from the pattern to using `engine = var.engine` in the `local`.
    1. Allow `var.redis_clusters` to supply `engine` with a default for `redis` and pass to module as-is
    1. Same as option 2 but allow `local.cluster_attributes` to overwrite it
        - This might be best of both worlds however no other argument does this so it would be breaking the pattern
- Went with option 2 so the argument isn't a new requirement for everyone and it doesn't break consistency

</details>

## references

- https://github.com/cloudposse/terraform-aws-elasticache-redis/releases/tag/v1.7.0
- https://aws.amazon.com/blogs/opensource/why-aws-supports-valkey/
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group#engine
- https://github.com/hashicorp/terraform-provider-aws/issues/39641
- https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.73.0
- https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/Clusters.Create.html
- https://www.lastweekinaws.com/blog/aws-valkey-play-when-a-fork-becomes-a-price-cut/

### DIFF
--- a/modules/elasticache-redis/modules/redis_cluster/variables.tf
+++ b/modules/elasticache-redis/modules/redis_cluster/variables.tf
@@ -14,12 +14,12 @@ variable "create_parameter_group" {
 variable "engine" {
   type        = string
   default     = "redis"
-  description = "Name of the cache engine"
+  description = "Name of the cache engine to use: either `redis` or `valkey`"
 }
 
 variable "engine_version" {
   type        = string
-  description = "Redis Version"
+  description = "Version of the cache engine to use"
   default     = "6.0.5"
 }
 
--- a/modules/elasticache-redis/modules/redis_cluster/variables.tf
+++ b/modules/elasticache-redis/modules/redis_cluster/variables.tf
@@ -11,6 +11,12 @@ variable "create_parameter_group" {
   description = "Whether new parameter group should be created. Set to false if you want to use existing parameter group"
 }
 
+variable "engine" {
+  type        = string
+  default     = "redis"
+  description = "Name of the cache engine"
+}
+
 variable "engine_version" {
   type        = string
   description = "Redis Version"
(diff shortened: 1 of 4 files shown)


GabisCampana:
@Nuru

modules/elasticache-redis/modules/redis_cluster/variables.tf
```diff
@@ -11,6 +11,12 @@ variable "create_parameter_group" {
   description = "Whether new parameter group should be created. Set to false if you want to use existing parameter group"
 }
 
+variable "engine" {
+  type        = string
+  default     = "redis"
+  description = "Name of the cache engine"
```
Nuru, line 17:
```suggestion
  description = "Name of the cache engine to use: either `redis` or `valkey`"
```

modules/elasticache-redis/modules/redis_cluster/variables.tf
```diff
@@ -11,6 +11,12 @@ variable "create_parameter_group" {
...
+}
+
 variable "engine_version" {
   type        = string
   description = "Redis Version"
```
Nuru, line 22:
```suggestion
  description = "Version of the cache engine to use"
```

Nuru (changes requested):
Please enumerate the legal values of `engine` in the variable description. Please also provide, in the README, an explanation of what they mean, with one or more links to even more information.

nitrocode:
@Nuru changes addressed, please re-review.

Nuru (approved):

PR #1171, opened 2024-10-30
feat(elasticache-redis): add snapshot retention limit

nitrocode:
## what

- add snapshot retention limit

## why

- Resolves [ElastiCache.1](https://docs.aws.amazon.com/securityhub/latest/userguide/elasticache-controls.html#elasticache-1) (ElastiCache (Redis OSS) clusters should have automatic backups enabled) securityhub control by giving the ability to set this value to a number greater than 0

## references

- Default is 0 https://github.com/cloudposse/terraform-aws-elasticache-redis/blob/9104d9a6a120ae9c90f59c5eb4ea711cc2d2c6bb/variables.tf#L223-L227
- Module received the feature 5 years ago so no need to update the module version https://github.com/cloudposse/terraform-aws-elasticache-redis/pull/45
- Related PR #1170 which upgrades the module


GabisCampana:
@Nuru

modules/elasticache-redis/modules/redis_cluster/variables.tf
```diff
@@ -62,6 +62,7 @@ variable "cluster_attributes" {
     automatic_failover_enabled      = bool
     auto_minor_version_upgrade      = bool
     auth_token_enabled              = bool
+    snapshot_retention_limit        = bool
```
Nuru, line 65:
```suggestion
    snapshot_retention_limit        = number
```

Nuru (changes requested):

modules/elasticache-redis/modules/redis_cluster/variables.tf
nitrocode: Good catch! Thank you

Nuru (approved):

PR #1205, opened 2026-01-14
docs: Adding vpc-routes component

ivan-pinatti:
# WHAT
- Added the `vpc-routes` component to the list

# WHY
- Component was missing


ivan-pinatti:
@goruha / @osterman please reconsider having the list updated.

It is a much cleaner and organized way to find the components. The new page doesn't list them and search is not that great.

goruha:
@osterman what do you think about this idea? ^

